PR #3, opened 2023-03-31
Develop

IvanGLS:


crud.py
```diff
@@ -0,0 +1,54 @@
...
+def get_author_by_name(db: Session, name: str) -> models.Author:
+    return db.query(models.Author).filter(models.Author.name == name).first()
+
+
+def get_book(db: Session, book_id: int) -> models.Book:
+    return db.query(models.Book).filter(models.Book.id == book_id).first()
+
+
+def get_books(db: Session, skip: int = 0, limit: int = 100) -> list[models.Book]:
+    return db.query(models.Book).offset(skip).limit(limit).all()
```
olena-hul: Improvement (fix only if you have a desire)
Create a BookService class to enforce some architecture in your project
Move crud methods inside this class
`db` should be instance attribute, and this way you won't need to pass it to every crud method

olena-hul (changes requested):

Bohdan-Salamakha (approved):

PR #7, opened 2023-04-03
Solution

MarharytaSovpenko:


library/main.py
```diff
@@ -0,0 +1,74 @@
...
+from sqlalchemy.orm import Session
+
+import crud
+import schemas
+from database import SessionLocal
+
+app = FastAPI()
+
+
+def get_db() -> Session:
```
olena-hul, line 11: Improvement (fix only if you have a desire)
Create a BookService class to enforce some architecture in your project
Move crud methods inside this class
db should be instance attribute, and this way you won't need to pass it to every crud method

olena-hul (changes requested):

PR #8, opened 2023-04-04
Solution

manjustice:


kuznetsb (approved):
LGTM

crud.py
```diff
@@ -0,0 +1,54 @@
+from sqlalchemy.orm import Session
+
+import models
+import schemas
+
+
+def get_authors(db: Session, skip: int = 0, limit: int = 50):
```
YuriiKindrat: also if you use annotations, use them for return types as well

main.py
```diff
@@ -0,0 +1,59 @@
...
+@app.get("/authors/", response_model=list[schemas.Author])
+def read_authors(
+        skip: int = 0,
+        limit: int = 50,
+        db: Session = Depends(get_db)
+):
+    return crud.get_authors(db=db, skip=skip, limit=limit)
+
+
+@app.get("/authors/{author_id}", response_model=schemas.Author)
```

YuriiKindrat (changes requested):

crud.py
YuriiKindrat: u didn't fix that here)

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #15, opened 2023-04-10
Solution

artemkazakov947:


alembic/versions/133cafaf47e9_added_field_books_to_author_model_.py
ArsenPidhoretskyi, line 1: Empty migrations

alembic/versions/8d69ab31325d_added_field_books_to_author_model.py
ArsenPidhoretskyi, line 1: Empty migration

alembic/versions/f031f82c91b3_innitial_migration.py
ArsenPidhoretskyi, line 1: Empty migration

crud.py
```diff
@@ -0,0 +1,46 @@
...
+        name=author.name,
+        bio=author.bio,
+    )
+    db.add(new_author)
+    db.commit()
+    db.refresh(new_author)
+    return new_author
+
+
+def get_author_list(db: Session):
```
ArsenPidhoretskyi: you can accept offset and limit here

ArsenPidhoretskyi (changes requested):
Add return type annotations

main.py
```diff
@@ -0,0 +1,63 @@
...
+@library.post("/author-create/", response_model=schemas.AuthorRead)
+def create_author(author: schemas.AuthorCreate, db: Session = Depends(get_db)) -> schemas.AuthorRead:
+    return crud.create_author(author=author, db=db)
+
+
+@library.get("/authors/{author_id}", response_model=schemas.AuthorRead)
+def get_author(
+        author_id: int,
+        db: Session = Depends(get_db),
+) -> HTTPException | schemas.AuthorRead:
```
olena-hul: HTTPException is not a value that is returned, it is an exception that is raised, so this can't be used as a return annotation 
We can include the info about exceptions in docstring, but it is not a good practice to include them in annotations

olena-hul (changes requested):

olena-hul (approved):

olena-hul:
Good job!

PR #22, opened 2023-04-30
Solution

MaybeTI:


DavidGlusch (approved):

crud.py
```diff
@@ -0,0 +1,116 @@
...
+
+from sqlalchemy.orm import Session
+
+import schemas
+from db import models
+from schemas import AuthorCreate, BookCreate
+from sqlalchemy import func, desc
+
+
+def get_all_books(
```
YuriiKindrat, line 11: if you use annotations, use them for return types as well

YuriiKindrat (changes requested):

crud.py
YuriiKindrat: what about other funcs ? )

YuriiKindrat (changes requested):
remove db from ur PR

PR #24, opened 2023-05-02
solution

DavidGlusch:


vladzabolotnyi (approved):

SafonovVladimir (approved):
GJ!

alembic/README
```diff
@@ -0,0 +1 @@
+Generic single-database configuration.
```
YuriiKindrat: no blank line at the end of file

library/crud.py
```diff
@@ -0,0 +1,80 @@
+from sqlalchemy.orm import Session
+
+from library.models import DBAuthor, DBBook
+
+from library.schemas import AuthorCreate, BookCreate
+
+
+def get_author_by_name(db: Session, author_name: str):
```
YuriiKindrat: also if you use annotations, use them for return types as well

library/database.py
```diff
@@ -0,0 +1,13 @@
...
+
+SQLALCHEMY_DATABASE_URL = "sqlite:///./library.db"
+
+
+engine = create_engine(
+    SQLALCHEMY_DATABASE_URL, connect_args={"check_same_thread": False}
+)
+SessionLocal = sessionmaker(autocommit=False, autoflush=False, bind=engine)
+
+Base = declarative_base()
```
YuriiKindrat: no blank line here too

library/main.py
```diff
@@ -0,0 +1,97 @@
...
+    return db_author
+
+
+@app.get("/authors/", response_model=list[schemas.Author])
+def read_authors(skip: int = 0, limit: int = 100, db: Session = Depends(get_db)):
+    authors = crud.get_authors(db=db, skip=skip, limit=limit)
+    return authors
+
+
+@app.get("/authors/{author_id}", response_model=schemas.Author)
```
YuriiKindrat: I don't think it's crucial, but as a rule endpoints should end with /

YuriiKindrat (changes requested):

library.db
wQuelS, line 1: never leave .db files in your repo

library/main.py
```diff
@@ -0,0 +1,125 @@
+from fastapi import Depends, FastAPI, HTTPException
+from sqlalchemy.orm import Session
+
+from library import crud, models, schemas
```
wQuelS, line 4: do you really like to write those `models.` or `schemas.` prefixes every time?

library/main.py
```diff
@@ -0,0 +1,125 @@
...
+
+
+@app.post("/authors/", response_model=schemas.Author)
+def create_author(
+        author: schemas.AuthorCreate,
+        db: Session = Depends(get_db)
+) -> models.DBAuthor:
+    db_author_name = crud.get_author_by_name(db=db, author_name=author.name)
+    if db_author_name:
+        raise HTTPException(status_code=404, detail="Author already exists")
```
wQuelS, line 26: I will not reject because of this, but now you know that such logic never appears at endpoints)

library/main.py
```diff
@@ -0,0 +1,125 @@
...
+@app.put("/authors/{author_id}/", response_model=schemas.Author)
+def update_author(
+        author_id: int,
+        author: schemas.AuthorCreate,
+        db: Session = Depends(get_db)
+) -> models.DBAuthor:
+    db_author = crud.update_author(db=db, author_id=author_id, author=author)
+    if db_author is None:
+        raise HTTPException(status_code=404, detail="Author not found")
+    return db_author
```
wQuelS, line 61: seems like you would face an error even before your validation in this method. Cuz in  `crud.update` the `get_author` can return `None` if there is no such author and  `    db_author.name = author.name` line will raise 500. Am I right?

library/main.py
```diff
@@ -0,0 +1,125 @@
...
+
+@app.delete("/authors/{author_id}/", response_model=schemas.Author)
+def delete_author(
+        author_id: int,
+        db: Session = Depends(get_db)
+) -> models.DBAuthor:
+    db_author = crud.delete_author(db=db, author_id=author_id)
+    if db_author is None:
+        raise HTTPException(status_code=404, detail="Author not found")
+    return db_author
```
wQuelS, line 72: do you really need to return something after delete? It is up to you btw, just asking)

library/models.py
```diff
@@ -0,0 +1,24 @@
...
+    books = relationship("DBBook", backref="author", lazy=True)
+
+
+class DBBook(Base):
+    __tablename__ = "book"
+
+    id = Column(Integer, primary_key=True)
+    title = Column(String(100), nullable=False)
+    summary = Column(String(500))
+    publication_date = Column(Date, nullable=False)
```
wQuelS, line 23: so it is possible to set any publication_date, even if it is future?

library/schemas.py
```diff
@@ -0,0 +1,47 @@
+from typing import Optional
+from pydantic import BaseModel
+from datetime import date
+
+
+class AuthorBase(BaseModel):
+    name: str
+    bio: Optional[str] = None
```
wQuelS, line 8: do you even use it for serialization?

library/schemas.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+class Author(AuthorBase):
+    id: int
+
+    class Config:
+        orm_mode = True
+
+
+class AuthorList(BaseModel):
```
wQuelS, line 22: and this

library/schemas.py
```diff
@@ -0,0 +1,47 @@
...
+    items: list[Author]
+    total: int
+
+
+class BookBase(BaseModel):
+    title: str
+    summary: Optional[str] = None
+    publication_date: date
+    author_id: int
```
wQuelS, line 31: and this

library/schemas.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+class Book(BookBase):
+    id: int
+
+    class Config:
+        orm_mode = True
+
+
+class BookList(BaseModel):
```
wQuelS, line 45: same

wQuelS (changes requested):

PR #27, opened 2023-05-02
Solution

yura-hudzovskyi:


Morty67 (approved):

crud.py
```diff
@@ -0,0 +1,151 @@
...
+
+from sqlalchemy import select
+from sqlalchemy.ext.asyncio import AsyncSession
+
+import models
+import schemas
+from sqlalchemy.orm import selectinload
+
+
+async def get_authors(db: AsyncSession, skip: int = 0, limit: int = 5):
```

main.py
```diff
@@ -0,0 +1,148 @@
...
+    tags=["auth"],
+)
+app.include_router(
+    fastapi_users.get_users_router(schemas.UserRead, schemas.UserUpdate),
+    prefix="/users",
+    tags=["users"],
+)
+
+
+@app.get("/authors", response_model=list[schemas.Author])
```

users.py
```diff
@@ -0,0 +1,55 @@
...
+
+auth_backend = AuthenticationBackend(
+    name="jwt",
+    transport=bearer_transport,
+    get_strategy=get_jwt_strategy,
+)
+
+fastapi_users = FastAPIUsers[User, uuid.UUID](get_user_manager, [auth_backend])
+
+current_active_user = fastapi_users.current_user(active=True)
```
YuriiKindrat: no blank line at the end of file

YuriiKindrat (changes requested):

crud.py
```diff
@@ -0,0 +1,154 @@
...
+            {
+                "id": book.id,
+                "title": book.title,
+                "summary": book.summary,
+                "publication_date": book.publication_date,
+                "author_id": book.author_id,
+            }
+            for book in author.books
+        ],
+    }
```
wQuelS, line 65:
```suggestion
    return models.Author.from_orm(author)
```
have you tried this ?

main.py
```diff
@@ -0,0 +1,148 @@
...
+)
+app.include_router(
+    fastapi_users.get_users_router(schemas.UserRead, schemas.UserUpdate),
+    prefix="/users",
+    tags=["users"],
+)
+
+
+@app.get("/authors/", response_model=list[schemas.Author])
+async def get_authors(db: AsyncSession = Depends(get_async_session)):
```
wQuelS, line 40: if you look carefully u'll mention that your endpoints don't use session at all as they don't perform any actions on db

main.py
```diff
@@ -0,0 +1,148 @@
...
+@app.post("/authors/create/", response_model=schemas.Author)
+async def create_author(
+    author: schemas.AuthorCreate,
+    db: AsyncSession = Depends(get_async_session),
+    current_user=Depends(current_active_user),
+):
+    if current_user:
+        return await crud.create_author(db=db, author=author)
+    else:
+        raise HTTPException(status_code=400, detail="Not enough permissions")
```
wQuelS, line 58: first - you don't need `else` statement after `return`, cuz it will reach line57 only in case `if current_user is False` . In all other cases, the function will stop on line 56.

second - it would be better to `if not current_user` -> throws an Exception and return in all other cases.

main.py
```diff
@@ -0,0 +1,148 @@
...
+):
+    if current_user:
+        return await crud.update_author(
+            db=db,
+            author_id=author_id,
+            name=name,
+            bio=bio,
+        )
+    else:
+        raise HTTPException(status_code=400, detail="Not enough permissions")
```
wQuelS, line 77: same

crud.py
```diff
@@ -0,0 +1,154 @@
...
+import schemas
+from sqlalchemy.orm import selectinload
+
+from models import Author, Book
+
+
+async def get_authors(db: AsyncSession, skip: int = 0, limit: int = 5) -> list[dict[str, list[dict[str, Any]] | Any]]:
+    stmt = (
+        select(models.Author)
+        .options(selectinload(models.Author.books))
```
wQuelS, line 17: you've already specified `lazy='selectin'` in your `Author's relationship`. It will perform `selecting` loading strategy on EVERY Author query for the `books` relationship, due to that `lazy` param

crud.py
```diff
@@ -0,0 +1,154 @@
...
+            ],
+        }
+        for author in authors
+    ]
+
+
+async def get_author(db: AsyncSession, author_id: int) -> dict:
+    stmt = (
+        select(models.Author)
+        .options(selectinload(models.Author.books))
```
wQuelS, line 46: same. SQLalchemy handles this and it will not cause any problems, but technically here you "override" selecting loading strat. with another selectinload

alembic/versions/882adca3cde9_init.py
wQuelS, line 1: usually `init` is the first migration. try to give short but descriptive names for each next migration

schemas.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+class UserUpdate(schemas.BaseUserUpdate):
+    pass
+
+
+class BookBase(BaseModel):
+    title: str
+    summary: str = None
+    publication_date: date
```
wQuelS, line 26: do you even use it ?

schemas.py
```diff
@@ -0,0 +1,59 @@
...
+    id: int
+    author_id: int
+
+    class Config:
+        orm_mode = True
+
+
+class AuthorBase(BaseModel):
+    name: str
+    bio: str = None
```
wQuelS, line 43: same

schemas.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+class AuthorCreate(AuthorBase):
+    pass
+
+
+class AuthorUpdate(AuthorBase):
+    name: str = None
```
wQuelS, line 51: and even this?

wQuelS (changes requested):

PR #30, opened 2023-05-02
'Solution'

Morty67:


SergiiMachulin (approved):
Not bad idea - to add annotations, and new lines at the end of some files. 
Also, why do both delete functions return Response 200, but not 204?
And you added to gitignore some files from alembic, but they are in your repo)

Morty67:
> Not bad idea - to add annotations, and new lines at the end of some files. Also, why do both delete functions return Response 200, but not 204? And you added to gitignore some files from alembic, but they are in your repo)

I am also surprised that I added some files to gitignore but they still got pushed.

.gitignore
```diff
@@ -128,3 +128,5 @@ dmypy.json
 # Pyre type checker
 .pyre/
 .idea/
+alembic/README
+alembic/script.py.mako
```
YuriiKindrat: no blank line at the end of file, fix everywhere

library_api/crud.py
```diff
@@ -0,0 +1,64 @@
+from sqlalchemy.orm import Session
+
+from library_api.db.models import DBAuthor, DBBook
+from library_api.schemas import AuthorCreate, BookCreate
+
+
+def get_all_authors(db: Session, skip: int = 0, limit: int = 50):
```
YuriiKindrat: also if you use annotations, use them for return types as well

library_api/main.py
```diff
@@ -0,0 +1,112 @@
...
+
+@app.get("/authors/", response_model=list[Author])
+def read_authors(
+    skip: int = 0, limit: int = 100, db: Session = Depends(get_db)
+):
+    authors = crud.get_all_authors(db, skip=skip, limit=limit)
+    return authors
+
+
+@app.get("/authors/{id}", response_model=Author)
```
YuriiKindrat: I don't think it's crucial, but as a rule endpoints should end with /

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #31, opened 2023-05-02
Solution

MikhailLyvak:


DmytryTong (approved):
I've been redirected from parsing theme first task.. Strange stuff, btw

crud.py
```diff
@@ -0,0 +1,77 @@
+from sqlalchemy.orm import Session
+
+from db import models
+import schemas
+
+
+def get_all_authors(db: Session):
```

main.py
```diff
@@ -0,0 +1,117 @@
...
+):
+    items = crud.get_book_list(
+        db=db, title=title, summary=summary
+    )
+    total = len(items)
+    items = items[skip: skip + limit]
+    return BookListResponse(items=items, total=total, limit=limit, offset=skip)
+
+
+@app.get("/book/{book_id}", response_model=schemas.Book)
```

YuriiKindrat (changes requested):
remove db from ur PR

crud.py
YuriiKindrat: u didn't fix that

YuriiKindrat (changes requested):

PR #32, opened 2023-05-02
Solution DonchykD

DmytryTong:


MikhailLyvak (changes requested):
We had to add paggination. )

crud.py
```diff
@@ -0,0 +1,62 @@
+from sqlalchemy.orm import Session
+
+import models
+import schemas
+
+
+def get_all_authors(db: Session, skip: int = 0, limit: int = 10):
```

YuriiKindrat (changes requested):

crud.py
DmytryTong: Fixed

alembic/README
```diff
@@ -0,0 +1 @@
+Generic single-database configuration.
```
wQuelS: add new line)

main.py
```diff
@@ -0,0 +1,86 @@
...
+    db = SessionLocal()
+
+    try:
+        yield db
+    finally:
+        db.close()
+
+@app.get("/")
+def root() -> dict:
+    return {"message": "Hello world"}
```
wQuelS: do you need it?

main.py
```diff
@@ -0,0 +1,86 @@
...
+    author: schemas.AuthorCreate,
+    db: Session = Depends(get_db),
+):
+    db_author = crud.get_author_by_name(db=db, name=author.name)
+
+    if db_author:
+        raise HTTPException(
+            status_code=400,
+            detail="Such name for author already exist"
+        )
```
wQuelS, line 53: such logic should not be handled in endpoints. You may leave this logic here and i won't reject it, but FYI: Endpoints are only responsible for handling HTTP requests and responses. Business logic is usually located in other places that are responsible for high-level stuff

main.py
```diff
@@ -0,0 +1,86 @@
...
+
+    return crud.create_author(db=db, author=author)
+
+
+@app.get("/book/", response_model=list[schemas.Book])
+def read_book(
+    author_id: int | None = None,
+    db: Session = Depends(get_db)
```
wQuelS, line 63: I know that this task was implemented according to guide, but FYI: usually it is not good to carry session through all modules of your project. If you look carefully you'll mention that endpoints are not using it at all:)

Again - you may leave it as it is)

main.py
```diff
@@ -0,0 +1,86 @@
...
+
+    return db_book
+
+
+@app.post("/book/", response_model=schemas.Book)
+def create_book(book: schemas.BookCreate, db: Session = Depends(get_db)):
+    return crud.create_book(db=db, book=book)
+
+
+add_pagination(app)
```
wQuelS: this task is rejected anyway due to redundant `root()` func and blank line, so can you please provide screenshot of paginated response from your endpoint? Just wanna look at it, cuz I've never used this lib)

models.py
```diff
@@ -0,0 +1,24 @@
+from sqlalchemy import Column, Integer, String, Date, ForeignKey
+from sqlalchemy.orm import relationship
+
+from database import Base
+
+
+class DBAuthor(Base):
```
wQuelS: those names:)))

models.py
```diff
@@ -0,0 +1,24 @@
...
+    __tablename__ = "authors"
+
+    id = Column(Integer, primary_key=True, index=True)
+    name = Column(String(255), nullable=False, unique=True)
```
wQuelS, line 11: unique name? Is this because of task requirements or you've done it by yourself ?

wQuelS (changes requested):

models.py
DmytryTong: Yes, that condition were in a Readme. 
image

models.py
DmytryTong: Renamed)

alembic/README
DmytryTong: Done)

main.py
DmytryTong: Nah, deleted)

main.py
DmytryTong: Refactored that thing to more usual limit-offset (now it's more readeble)

DmytryTong: image

main.py
DmytryTong: 'Kay, will do better next time

main.py
DmytryTong: 'Kay, will do better next time

wQuelS (approved):

PR #33, opened 2023-05-02
Solution

WDemidenko:


ZhAlexR (approved):

Y-Havryliv (approved):
gj, but if you use annotations, use them everywhere. You didn't use them for return types here.

PR #35, opened 2023-05-02
Solution

oksana-feshchenko:


oksanaaam (approved):

alembic/README
```diff
@@ -0,0 +1 @@
+Generic single-database configuration.
```
YuriiKindrat, line 1: no blank line at the end of file

crud.py
```diff
@@ -0,0 +1,96 @@
+from sqlalchemy.orm import Session
+
+import models
+import schemas
+
+
+def get_all_authors(db: Session, skip: int = 0, limit: int = 100):
```
YuriiKindrat, line 7: what about type annotation for return type here? fix everywhere

main.py
```diff
@@ -0,0 +1,128 @@
...
+    db_author = crud.get_author_by_name(db=db, name=author.name)
+    print(db_author)
+    if db_author:
+        raise HTTPException(
+            status_code=400, detail="Author with this name already exists."
+        )
+    return crud.create_author(db=db, author=author)
+
+
+@app.get("/authors/{author_id}", response_model=schemas.Author)
```
YuriiKindrat, line 37: I don't think it's crucial, but as a rule endpoints should end with /

YuriiKindrat (changes requested):

PR #36, opened 2023-05-03
Solution

ghost:


TeliksPG (approved):
good!

main.py
```diff
@@ -0,0 +1,97 @@
...
+    author_id: int,
+    db: Session = Depends(get_db),
+):
+    db_author = crud.get_author(db=db, author_id=author_id)
+    if db_author is None:
+        raise HTTPException(status_code=404, detail="Author not found")
+    return db_author
+
+
+# Books endpoints
```
Y-Havryliv, line 54: don't use comments like that, it doesn't give any additional info. It's just some statement.

Y-Havryliv (approved):
approved, gj, but your code should be as consistent as possible. If you used type annotations use them for return values as well.

PR #37, opened 2023-05-03
Solution

ZhAlexR:


ada-krav (approved):

crud.py
```diff
@@ -0,0 +1,53 @@
+from sqlalchemy.orm import Session
+
+from schemas import AuthorCreate, BookCreate
+
+from models import Author, Book
+
+
+def create_author(db: Session, author: AuthorCreate):
```

main.py
```diff
@@ -0,0 +1,68 @@
...
+    return crud.create_author(db=db, author=author)
+
+
+@app.get("/authors/", response_model=list[schemas.AuthorMain])
+def read_authors(limit: int = 50, db: Session = Depends(get_db)):
+    authors = crud.get_author_list(db=db, limit=limit)
+    return authors
+
+
+@app.get("/authors/{author_id}", response_model=schemas.AuthorMain)
```

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #38, opened 2023-05-03
Solution

inrock1:


oigod (approved):

crud.py
```diff
@@ -0,0 +1,65 @@
+from sqlalchemy.orm import Session
+
+from db import models
+import schemas
+
+
+def get_all_authors(db: Session):
```
YuriiKindrat: also if you use annotations, use them for return types as well

YuriiKindrat (changes requested):

crud.py
inrock1: annotations added

YuriiKindrat (approved):

PR #39, opened 2023-05-03
Fastapi

oksanaaam:


amber-marichi (approved):
i like it, probably could've added Author related schema to Book detail schema, like you did with list of Books inside Author schema. But its ok

crud.py
```diff
@@ -0,0 +1,44 @@
+from sqlalchemy.orm import Session
+from db import models
+
+import schemas
+
+
+def get_author(db: Session, author_id: int):
```

main.py
```diff
@@ -0,0 +1,66 @@
...
+    db = SessionLocal()
+    try:
+        yield db
+    finally:
+        db.close()
+
+
+@app.get("/")
+def read_root():
+    return {"Hello": "World"}
```

main.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+@app.get("/authors/", response_model=list[schemas.Author])
+def get_authors(skip: int = 0, limit: int = 10, db: Session = Depends(get_db)):
+    authors = crud.get_all_authors(db=db)
+    return authors[skip: skip + limit]
+
+
+@app.get("/authors/{author_id}", response_model=schemas.Author)
```

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #40, opened 2023-05-03
Solution

TeliksPG:


danylo-d (approved):

crud.py
```diff
@@ -0,0 +1,60 @@
+from sqlalchemy.orm import Session
+
+from db import models
+import schemas
+
+
+def get_all_authors(db: Session):
```

main.py
```diff
@@ -0,0 +1,69 @@
...
+    try:
+        yield db
+    finally:
+        db.close()
+
+
+@app.get("/")
+def root() -> dict:
+    return {"message": "Hello World"}
+
```

YuriiKindrat (changes requested):

PR #41, opened 2023-05-03
implement fastapi library project

oigod:


ykopatko (approved):
gj

main.py
```diff
@@ -0,0 +1,66 @@
...
+    return crud.create_author(db=db, author=author)
+
+
+@app.get("/authors/", response_model=list[schemas.Author])
+def read_authors(skip: int = 0, limit: int = 100, db: SessionLocal = Depends(get_db)):
+    authors = crud.get_all_authors(db, skip=skip, limit=limit)
+    return authors
+
+
+@app.get("/authors/{author_id}", response_model=schemas.Author)
```
YuriiKindrat, line 32: it is better to use the same codestyle as for urls before - add `/` to the end of url

YuriiKindrat (changes requested):

PR #42, opened 2023-05-03
Solution

ada-krav:


ostapT (approved):

crud.py
```diff
@@ -0,0 +1,73 @@
+from sqlalchemy.orm import Session
+from models import Author, Book
+from schemas import AuthorCreate, AuthorUpdate, BookCreate, BookUpdate
+
+
+def create_author(db: Session, author: AuthorCreate):
```
YuriiKindrat, line 6: if you use annotations, use them for return types as well

main.py
```diff
@@ -0,0 +1,59 @@
...
+def create_author_endpoint(author: AuthorCreate, db: Session = Depends(get_db)):
+    return create_author(db=db, author=author)
+
+
+@app.get("/authors/", response_model=List[Author])
+def get_authors_endpoint(skip: int = 0, limit: int = 10, db: Session = Depends(get_db)):
+    return get_authors(db=db, skip=skip, limit=limit)
+
+
+@app.get("/authors/{author_id}", response_model=Author)
```
YuriiKindrat, line 34: I don't think it's crucial, but as a rule endpoints should end with /

YuriiKindrat (changes requested):
what about alembic files ?

PR #44, opened 2023-05-03
Solution

ostapT:


crud.py
```diff
@@ -0,0 +1,73 @@
+from sqlalchemy.orm import Session
+
+import models
+import schemas
+from models import DBAuthor
+
+
+def get_all_authors(db: Session,  skip: int = 0, limit: int = 10):
```
YuriiKindrat, line 8: if you use annotations, use them for return types as well

main.py
```diff
@@ -0,0 +1,82 @@
...
+    try:
+        yield db
+    finally:
+        db.close()
+
+
+@app.get("/")
+def root():
+    return {"message": "HELLO!"}
+
```
YuriiKindrat, line 23: why do u need this ?)

main.py
```diff
@@ -0,0 +1,82 @@
...
+    if db_author is None:
+        raise HTTPException(
+            status_code=404,
+            detail="Author not found"
+        )
+
+    return db_author
+
+
+@app.get("/books", response_model=list[schemas.Book])
```
YuriiKindrat, line 59: use the same style for urls as before - add `/` to the end of urls

YuriiKindrat (changes requested):

PR #45, opened 2023-05-03
Solution

danylo-d:


main.py
```diff
@@ -0,0 +1,51 @@
...
+    finally:
+        db.close()
+
+
+@app.get("/authors/", response_model=list[schemas.Author])
+def get_authors(db: SessionLocal = Depends(get_db)):
+    return crud.get_authors(db)
+
+
+@app.get("/authors/{author_id}", response_model=schemas.Author)
```
YuriiKindrat, line 24: I don't think it's crucial, but as a rule endpoints should end with /

crud.py
```diff
@@ -0,0 +1,59 @@
+from sqlalchemy.orm import Session
+
+import models
+import schemas
+
+
+def get_authors(db: Session):
```
YuriiKindrat, line 7: what about type annotations for return types ?

alembic/README
```diff
@@ -0,0 +1 @@
+Generic single-database configuration.
```
YuriiKindrat, line 1: no blank line at the end of file

YuriiKindrat (changes requested):

PR #46, opened 2023-05-03
Solution

Litvinenko23:


NEkropo1 (approved):
Looks better than my code

crud.py
```diff
@@ -0,0 +1,71 @@
+from typing import List, Optional
+
+from sqlalchemy.orm import Session
+
+from db import models
+import schemas
+
+
+def get_all_books(db: Session):
```

YuriiKindrat (changes requested):
+

YuriiKindrat (changes requested):
what about alembic files ?

YuriiKindrat (approved):

PR #47, opened 2023-05-03
Solution

ykopatko:


crud.py
```diff
@@ -0,0 +1,64 @@
+from sqlalchemy.orm import Session
+
+from models import DBAuthor, DBBook
+from schemas import AuthorCreate, BookCreate, BookCreateForAuthor
+
+
+def get_all_authors(db: Session, skip: int = 0, limit: int = 10):
```

main.py
```diff
@@ -0,0 +1,68 @@
...
+    finally:
+        db.close()
+
+
+@app.get("/authors/", response_model=list[Author])
+def get_authors(db: Session = Depends(get_db)):
+    return crud.get_all_authors(db=db)
+
+
+@app.get("/authors/{author_id}", response_model=Author)
```

YuriiKindrat (changes requested):

crud.py
YuriiKindrat: u didn't fix that here )

YuriiKindrat (changes requested):

Serhii-Leonenko (approved):

PR #48, opened 2023-05-03
Solution

Maytreyaya:


crud.py
```diff
@@ -0,0 +1,48 @@
+from sqlalchemy.orm import Session
+from db import models
+import schemas
+
+
+def get_all_books(db: Session, skip: int = 0, limit: int = 100):
```

main.py
```diff
@@ -0,0 +1,72 @@
...
+    db = SessionLocal()
+    try:
+        yield db
+    finally:
+        db.close()
+
+
+@app.get("/")
+def root():
+    return {"message": "Hallo world"}
```

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #49, opened 2023-05-03
Learn flow

NEkropo1:


crud.py
```diff
@@ -0,0 +1,63 @@
+from fastapi import HTTPException
+from sqlalchemy.orm import Session
+
+import models
+import schemas
+
+
+def get_all_authors(db: Session):
```

main.py
```diff
@@ -0,0 +1,71 @@
...
+def read_books(
+        skip: int = 0,
+        limit: int = 100,
+        author_id: int | None = None,
+        db: Session = Depends(get_db)
+):
+    return crud.get_all_books(db, author_id=author_id)
+
+
+@app.get("/books/{book_id}", response_model=schemas.Book)
```

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #50, opened 2023-05-03
solution (except of cascade deletion)

Yulia-Mala:


crud.py
```diff
@@ -0,0 +1,62 @@
+from fastapi import HTTPException, Response
+from sqlalchemy.orm import Session
+from starlette import status
+
+from models import DBAuthor, DBBook
+from schemas import BookNoId
+
+
+def get_many(db_session: Session, db_model):
```

YuriiKindrat (changes requested):

crud.py
```diff
@@ -0,0 +1,64 @@
...
+    db_session.refresh(new_item)
+
+    return new_item
+
+
+def update_item(ident: int, db_session: Session, update_data, db_model: Type) -> Base:
+    db_session.query(db_model).filter(db_model.id == ident).update(update_data.dict())
+    db_session.commit()
+
+    return db_session.get(db_model, ident)
```
wQuelS, line 33: If there is no record in db with such `ident` this func will return `None`. Since your update endpoints have `reposnse_model` validation - FastAPI will try to serialize that `None` on pydantic schema provided. This will cause 500 error which is not good

alembic/README
```diff
@@ -0,0 +1 @@
+Generic single-database configuration.
```
wQuelS, line 1: blank line :)

crud.py
```diff
@@ -0,0 +1,64 @@
...
+from models import DBAuthor, DBBook
+from schemas import BookNoId
+
+
+def get_many(db_session: Session, db_model: Type) -> list:
+    return db_session.query(db_model).all()
+
+
+def get_one(ident: int, db_session: Session, db_model: Type) -> Base:
+    return db_session.get(db_model, ident)
```
wQuelS, line 16: It can return `None` and cause 500 error when FastAPI will try to serialize it before giving response

models.py
```diff
@@ -0,0 +1,23 @@
...
+from database import Base
+
+
+class DBAuthor(Base):
+    __tablename__ = "author"
+
+    id = Column(Integer, primary_key=True, index=True)
+    name = Column(String(127), unique=True, nullable=False)
+    bio = Column(String(511), nullable=True)
+
```
wQuelS, line 13: You did not include field for relation with book, and this will not cause some errors or whatever. BUT
You will not have backward relation and when you load the author, you will not be able to `author.books`. To retrieve all books written by some author - you'll have to add additional query by yourself.

And also, task requirement asked you to do that:)

models.py
```diff
@@ -0,0 +1,23 @@
...
+
+class DBBook(Base):
+    __tablename__ = "book"
+
+    id = Column(Integer, primary_key=True, index=True)
+    title = Column(String(255), nullable=False)
+    summary = Column(String(1023), nullable=True)
+    publication_date = Column(DateTime(timezone=True), server_default=func.now())
```
wQuelS, line 21: if this field means the time when book was published - then it would be pretty strange that book is published at the same time it was added to your system:)
And imagine situation: in the future, you'll want to add some logic to `bulk_add` which basically will add many books to your system at once. It would be weird if an author will have 10-15 books with the same publication date.

schemas.py
```diff
@@ -0,0 +1,33 @@
...
+    id: int
+
+
+class Book(BaseModel):
+    title: str
+    summary: str
+    publication_date: date
+
+    class Config:
+        orm_mode = True
```
wQuelS, line 24: do you even use this schema? Usage as a schema to inherit from does not count here in this case
Also:
Generally, you should not use `orm_mode=True` like this. I mean that this option should be set on demand in places that really need it. Because with this option pydantic will use less efficient algorithm to process date upon validation and so on. 
Only if you know for sure that all schemas need this options - your approach is ok

main.py
```diff
@@ -0,0 +1,62 @@
...
+from crud import get_many, get_one, update_item, get_books, add_book, add_item, delete_item
+from database import get_session
+from models import DBAuthor, DBBook
+from schemas import AuthorWithId, AuthorNoId, BookWithId, BookNoId
+
+app = FastAPI(title="Library Manager")
+
+
+@app.get("/authors/", response_model=Page[AuthorWithId])
+def read_authors(db_session: Session = Depends(get_session)):
```
wQuelS, line 13: your endpoints don't use `session` at all. so maybe they should not even know about session?

main.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+@app.delete("/authors/{ident:int}/")
+def delete_author(ident, db_session: Session = Depends(get_session)):
+    return delete_item(ident=ident, db_session=db_session, db_model=DBAuthor)
+
+
+@app.get("/books/", response_model=Page[BookWithId])
+def read_books(db_session: Session = Depends(get_session), author_ids: str | None = None):
+    return paginate(get_books(db_session=db_session, author_ids=author_ids))
```
wQuelS, line 39: have you tried this endpoint without providing `author_ids`?

main.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+@app.post("/books/", response_model=BookNoId)
+def create_book(book_data: BookNoId, db_session: Session = Depends(get_session)):
+    return add_book(db_session=db_session, book_data=book_data)
+
+
+@app.patch("/books/{ident:int}/", response_model=BookNoId)
+def update_book(ident, update_data: BookNoId, db_session: Session = Depends(get_session)):
+    return update_item(ident=ident, update_data=update_data, db_session=db_session, db_model=DBBook)
```
wQuelS, line 54: is it really `patch`? Cuz at this endpoint it is not possible to partially update the book, only "full-upd"

crud.py
```diff
@@ -0,0 +1,64 @@
...
+    new_item = db_model(**item_data.dict())
+
+    db_session.add(new_item)
+    db_session.commit()
+    db_session.refresh(new_item)
+
+    return new_item
+
+
+def update_item(ident: int, db_session: Session, update_data, db_model: Type) -> Base:
```
wQuelS, line 29: type hint for `update_data` ?

main.py
```diff
@@ -0,0 +1,62 @@
...
+
+@app.get("/books/{ident:int}/", response_model=BookWithId)
+def read_book(ident, db_session: Session = Depends(get_session)):
+    return get_one(ident=ident, db_session=db_session, db_model=DBBook)
+
+
+@app.post("/books/", response_model=BookNoId)
+def create_book(book_data: BookNoId, db_session: Session = Depends(get_session)):
+    return add_book(db_session=db_session, book_data=book_data)
+
```
wQuelS, line 50: so upon creation, books don't have authors that have written them? Seems like not really good design. But maybe you have some specific idea, so feel free to debate

wQuelS (changes requested):
Nice that your default crud functions do not depend on the model! And really good work in general.
Left some comments, tho)

PR #51, opened 2023-05-04
Develo

amber-marichi:


alembic/README
```diff
@@ -0,0 +1 @@
+Generic single-database configuration.
```
YuriiKindrat: no blank line at the end of file

main.py
```diff
@@ -0,0 +1,183 @@
...
+
+import crud, schemas, utils
+
+
+app = FastAPI()
+
+
+@app.get("/")
+def root():
+    return {"Hello": "World"}
```

main.py
```diff
@@ -0,0 +1,183 @@
...
+):
+    return crud.get_author_list(
+        db, skip=skip,
+        limit=limit,
+        author_name=author_name,
+        sort_field=sort_field,
+    )
+
+
+@app.get("/authors/{author_id}", response_model=schemas.Author)
```
YuriiKindrat: I don't think it's crucial, but as a rule endpoints should end with /, fix this everywhere

YuriiKindrat (changes requested):

main.py
amber-marichi: thanks, will do

main.py
amber-marichi: oops, used it when did alot of quicktests. will be removed

YuriiKindrat (approved):

utils.py
```diff
@@ -0,0 +1,101 @@
...
+                status_code = status.HTTP_401_UNAUTHORIZED,
+                detail="Token expired",
+            )
+        user = db.query(models.User).get(payload["id"])
+    except(JWTError, ValidationError):
+        raise HTTPException(
+            status_code=status.HTTP_401_UNAUTHORIZED,
+            detail="Invalid email of password"
+        )
+    return schemas.User.from_orm(user)
```
wQuelS: really nice that u've done so much additional stuff. But let's look closer at this func. Don't you think it has way too many responsibilities? Also, why this function is directly accessing database ?

models.py
```diff
@@ -0,0 +1,43 @@
...
+from datetime import datetime
+
+from database import Base
+
+
+class DBAuthor(Base):
+    __tablename__ = "authors"
+
+    id = Column(Integer, primary_key=True, index=True)
+    name = Column(String(255), unique=True, nullable=False)
```
wQuelS, line 19: unique name? Is it 'cuz of task requirements or u've decided to do so by yourself?

models.py
```diff
@@ -0,0 +1,43 @@
...
+    author = relationship("DBAuthor", back_populates="books")
+
+
+class User(Base):
+    __tablename__ = "users"
+
+    id = Column(Integer, primary_key=True, index=True)
+    email = Column(String, unique=True, index=True)
+    hashed_pass = Column(String)
+    date_created = Column(DateTime, default=datetime.utcnow)
```
wQuelS: you could use `server_default=now()`

models.py
```diff
@@ -0,0 +1,43 @@
...
+    books = relationship("DBBook", back_populates="author")
+
+
+class DBBook(Base):
+    __tablename__ = "books"
+
+    id = Column(Integer, primary_key=True, index=True)
+    title = Column(String(255), unique=True, nullable=False)
+    summary = Column(String(511), nullable=False)
+    publication_date = Column(Date)
```
wQuelS: if publication_date is the time when book is added to your system - then it should has some sort of automatic timestamp.
 If this field is just about when book was published itself - then I guess it is okay ( btw, it is possible to set this date to be in future), but maybe should have not nullable constraint

main.py
```diff
@@ -0,0 +1,178 @@
...
+app = FastAPI()
+
+
+@app.get("/authors/", response_model=list[schemas.Author])
+def read_authors(
+    skip: int = 0,
+    limit: int = 10,
+    author_name: str | None = None,
+    sort_field: str | None = None,
+    db: Session = Depends(utils.get_db)
```
wQuelS: why not import `get_db` func? It is annoying to write that `utils` prefix every time, isn't it?

crud.py
```diff
@@ -0,0 +1,164 @@
...
+    limit: int = 10,
+    author_name: str | None = None,
+    sort_field: str | None = None,
+):
+    queryset = db.query(models.DBAuthor)
+    if author_name is not None:
+        queryset = queryset.filter(models.DBAuthor.name.contains(author_name))
+    if sort_field is not None:
+        if sort_field.startswith("-"):
+            sort_field = sort_field[1:] + " desc"
```
wQuelS: one line ?

crud.py
```diff
@@ -0,0 +1,164 @@
...
+    sort_field: str | None = None
+):
+    queryset = db.query(models.DBBook)
+    if author_id is not None:
+        queryset = queryset.filter(models.DBBook.author_id == author_id)
+    if book_title is not None:
+        queryset = queryset.filter(models.DBBook.title.contains(book_title))
+    if sort_field is not None:
+        if sort_field.startswith("-"):
+            sort_field = sort_field[1:] + " desc"
```
wQuelS: same

crud.py
```diff
@@ -0,0 +1,164 @@
...
+    return db_author
+
+
+def delete_author(author_id: int, db: Session):
+    db_author = get_author(db, author_id=author_id)
+    if db_author is None:
+        raise HTTPException(
+            status_code=404,
+            detail="Author was not found"
+        )
```
wQuelS: seems like I've seen this piece of code before in you app:))
DRY

crud.py
```diff
@@ -0,0 +1,164 @@
...
+            status_code=404,
+            detail="Book was not found"
+        )
+    book_data = book.dict(exclude_unset=True)
+    for key, val in book_data.items():
+        setattr(db_book, key, val)
+    db.add(db_book)
+    db.commit()
+    db.refresh(db_book)
+    return db_book
```
wQuelS: many of your functions in this module have way too much responsibilities. They are checking for existence, handling errors, and perform db actions. Don't forget about SRP :)

schemas.py
```diff
@@ -0,0 +1,56 @@
...
+class AuthorBase(BaseModel):
+    name: str
+    bio: str | None = None
+
+    class Config:
+        orm_mode = True
+
+
+class BookCreate(BookBase):
+    author_id: int
```
wQuelS: what happens if i do not provide `summary` during book creation ?

schemas.py
```diff
@@ -0,0 +1,56 @@
...
+        orm_mode = True
+
+
+class AuthorBase(BaseModel):
+    name: str
+    bio: str | None = None
+
+    class Config:
+        orm_mode = True
+
```
wQuelS: is this class needed for all schemas that inherit from this one? If not - it is better to provide this option on demand. By using `orm_mode=True` pydantic will validate your data with a less efficient algorithm. You will not mention performance diff on such a simple API, but it is good to know:)

models.py
```diff
@@ -0,0 +1,43 @@
...
+    author_id = Column(Integer, ForeignKey("authors.id"))
+
+    author = relationship("DBAuthor", back_populates="books")
+
+
+class User(Base):
+    __tablename__ = "users"
+
+    id = Column(Integer, primary_key=True, index=True)
+    email = Column(String, unique=True, index=True)
```
wQuelS, line 41: uniquness on db lvl is nice, but what about format?

utils.py
```diff
@@ -0,0 +1,101 @@
...
+
+    return token
+
+
+def create_refresh_token(user: models.User):
+    user_obj = schemas.User.from_orm(user)
+    user_dict = user_obj.dict()
+    del user_dict["date_created"]
+    expire = datetime.utcnow() + timedelta(minutes=REFRESH_TOKEN_EXPIRE_MINUTES)
+    user_dict.update({"exp": expire})
```
wQuelS: it would be more appropriate to construct a new payload dictionary with just the data that you want to include in the token, rather than adding the 'exp' field to the user dictionary.

wQuelS (changes requested):
Nice work, left few comments.
Overall - you should think more about SRP principle

crud.py
amber-marichi: i see, so i just can move code with raising 404 into get_author func? ok ill try

crud.py
amber-marichi: yes, can do

models.py
amber-marichi: yes it was supposed to be added manually. I havnt thought it should be mandatory tbh. But i guess i'll try to make it not nullable and also use validator in schema. TY!

models.py
amber-marichi: it is "now" from sqlalchemy, right? ok ty

models.py
amber-marichi: author name should be unique according to the task, yeah.

schemas.py
amber-marichi: Thank you, noted! Indeed good to know, is it allright in such cases to added orm_mode=True for both read and write schemas, i mean it wont be SRP violation right? :)

models.py
amber-marichi: i was thinking about email validation. I thought its oke to use schema validator but i wasnt sure. What would you recommend ?

schemas.py
amber-marichi: It's going to be with null summary right?..  i thought it should be allright since it wasnt a mandatory field according to the task. But right now looking at my code im not sure. Will it be ok if i make it nullable for model field?

utils.py
amber-marichi: yep, my bad :) i definitely should've made it more readable and reuse get_user func. thanks

models.py
amber-marichi: idk maybe its related to SQLite but server_default was not working for me at all. :(

YuriiKindrat (approved):

PR #53, opened 2023-05-04
Develop

tytomyr:


crud.py
```diff
@@ -0,0 +1,79 @@
+from sqlalchemy.orm import Session
+from db.models import Author, Book
+from schemas import (AuthorCreate,
+                     AuthorUpdate,
+                     BookCreate)
+
+
+def create_author(
```
YuriiKindrat, line 8: if you use annotations, use them for return types as well

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #55, opened 2023-05-04
Solution

usrofgh:


crud.py
```diff
@@ -0,0 +1,96 @@
+from sqlalchemy.orm import Session, joinedload
+
+import schemas
+import models
+
+
+def get_author_list(
```

YuriiKindrat (changes requested):

alembic/README
```diff
@@ -0,0 +1 @@
+Generic single-database configuration.
```
wQuelS, line 1: new line:)

library.db
wQuelS, line 1: never leave .db files in your repo

main.py
```diff
@@ -0,0 +1,106 @@
+from fastapi import FastAPI, Depends, HTTPException
+from fastapi_pagination import add_pagination, Page, paginate
+from sqlalchemy.orm import Session
+from http import HTTPStatus
+
+import crud
+import schemas
```
wQuelS, line 7: do you really like to write those `schema` prefixes before every Pydantic model?

schemas.py
```diff
@@ -0,0 +1,37 @@
...
+
+from pydantic import BaseModel
+
+
+class AuthorBase(BaseModel):
+    name: str
+    bio: str
+
+    class Config:
+        orm_mode = True
```
wQuelS, line 11: Are u sure that every schema that inherits from this base class needs this option?
In general, you should use it on demand, since when `orm_mode=True` pydantic will use a less efficient algorithm when performing actions

schemas.py
```diff
@@ -0,0 +1,37 @@
...
+    pass
+
+
+class BookBase(BaseModel):
+    title: str
+    summary: str
+    publication_date: datetime.date
+    author_id: int
+
+    class Config:
```
wQuelS, line 28: same

models.py
```diff
@@ -0,0 +1,26 @@
+from sqlalchemy import Column, Integer, String, Text, ForeignKey, Date
+from sqlalchemy.orm import relationship
+
+from database import Base
+
+
+class DBAuthor(Base):
```
wQuelS, line 7: why do you need those weird names? You anyway don't import them directly, but using `models` prefix)

models.py
```diff
@@ -0,0 +1,26 @@
...
+    books = relationship("DBBook", back_populates="author")
+
+
+class DBBook(Base):
+    __tablename__ = "books"
+
+    id = Column(Integer, primary_key=True, index=True)
+    title = Column(String(255), unique=True)
+    summary = Column(Text)
+    publication_date = Column(Date)
```
wQuelS, line 23: so it is possible to set any date for publication, even if it is future?
Also, there can be no publication date at all, but with the existing author!?

wQuelS (changes requested):

PR #57, opened 2023-05-04
Solution

viktoria-rybenchuk:


crud.py
```diff
@@ -0,0 +1,76 @@
+from sqlalchemy.orm import Session
+import models
+import schemas
+from security import get_password_hash
+
+
+def get_author_list(
```

main.py
```diff
@@ -0,0 +1,125 @@
...
+def get_db() -> Session:
+    db = SessionLocal()
+
+    try:
+        yield db
+    finally:
+        db.close()
+
+
+@app.post("/token", response_model=Token)
```

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #58, opened 2023-05-08
Develop

Yevheniy-Lototskiy:


crud.py
```diff
@@ -0,0 +1,56 @@
+from sqlalchemy.orm import Session
+
+import schemas
+from db import models
+from schemas import AuthorCreate
+
+
+def get_all_authors(db: Session):
```

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #59, opened 2023-05-08
Dev

Anatolii-Poznyak:


crud.py
```diff
@@ -0,0 +1,75 @@
+from sqlalchemy.orm import Session
+
+import models
+import schemas
+
+
+def get_all_authors(db: Session, skip: int = 0, limit: int = 10):
```

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #60, opened 2023-05-08
Solution

me1nyk:


main.py
```diff
@@ -0,0 +1,66 @@
...
+    return author
+
+
+@app.get("/books", response_model=list[schemas.Book])
+def read_books(skip: int = 0, limit: int = 10, db: Session = Depends(get_db)):
+    books = crud.get_books(db, skip=skip, limit=limit)
+    return books
+
+
+@app.get("/books/{book_id}", response_model=schemas.Book)
```
YuriiKindrat, line 51: make the same style for urls as before, add `/` to the end of url

YuriiKindrat (changes requested):

PR #62, opened 2023-05-11
Develop

alexlega:


crud.py
```diff
@@ -0,0 +1,80 @@
+from sqlalchemy.orm import Session
+
+from models import Author, Book
+import schemas
+
+
+def create_author(db: Session, author: schemas.AuthorCreate):
```

main.py
```diff
@@ -0,0 +1,97 @@
...
+    return crud.create_author(db=db, author=author)
+
+
+@app.get("/authors/", response_model=List[schemas.Author])
+def read_authors(skip: int = 0, limit: int = 100, db: Session = Depends(get_db)):
+    authors = crud.get_authors(db, skip=skip, limit=limit)
+    return authors
+
+
+@app.get("/authors/{author_id}", response_model=schemas.Author)
```
YuriiKindrat, line 37: I don't think it's crucial, but as a rule endpoints should end with /

YuriiKindrat (changes requested):

PR #86, opened 2023-07-26
Solution v_1.0

bohdan-yatsyna:


crud.py
```diff
@@ -0,0 +1,69 @@
...
+    ).offset(skip).limit(limit).all()
+
+
+def get_books_list_by_author_id(
+    db: Session, author_id: int | None = None
+) -> list[models.Book]:
+    queryset = db.query(models.Book)
+
+    if author_id is not None:
+        return queryset.filter(models.Book.author_id == author_id).all()
```
ArsenPidhoretskyi:
```suggestion
        queryset = queryset.filter(models.Book.author_id == author_id)
```
Better to do like this. It will help to support code

main.py
```diff
@@ -0,0 +1,91 @@
+from fastapi import FastAPI, Depends, HTTPException
+from sqlalchemy.orm import Session
+
+import crud
+import models
+import schemas
+from database import SessionLocal
+
+app = FastAPI()
```
ArsenPidhoretskyi, line 11: image

crud.py
```diff
@@ -0,0 +1,69 @@
...
+def get_author(db: Session, author_id: int) -> models.Author | None:
+    queryset = db.query(models.Author)
+
+    return queryset.filter(models.Author.id == author_id).first()
+
+
+def get_authors(
+        db: Session,
+        skip: int = 0,
+        limit: int = 33,
```
ArsenPidhoretskyi: Why limit is 33?)

ArsenPidhoretskyi: As for me better to move default limit to separate constant and reuse it in all places

ArsenPidhoretskyi (changes requested):

crud.py
bohdan-yatsyna: All your remarks noted with thanks and the code is fixed accordingly.

vsmutok (approved):
Good Job 👍

PR #89, opened 2023-08-02
Solution

MkrtychKhachatrian:


schemas.py
```diff
@@ -0,0 +1,38 @@
...
+    author_id: int
+
+
+class BookList(BookBase):
+    id: int
+    author_id: int
+
+    class Config:
+        orm_mode = True
+
```
ArsenPidhoretskyi, line 38: This line must be empty check code

crud.py
```diff
@@ -0,0 +1,61 @@
...
+
+from sqlalchemy.orm import Session
+
+import models
+import schemas
+
+
+def get_author(db: Session, author_id: int) -> models.Author | None:
+    return db.query(models.Author
+                    ).filter(models.Author.id == author_id).first()
```
ArsenPidhoretskyi, line 11:
```suggestion
    return db.query(models.Author).filter(
        models.Author.id == author_id
    ).first()
```

ArsenPidhoretskyi (changes requested):

PR #112, opened 2023-08-27
Solution

Kasatello:


Andriy-Sydorenko (approved):
Good job, but it's better to move crud.py and schemas.py out of the db directory, into your project folder, or maybe even into a separate app package

crud.py
Oleg-s-s, line 1: image

Oleg-s-s (changes requested):

PR #127, opened 2023-10-26
Solution

artur-leleiko:


crud.py
```diff
@@ -0,0 +1,46 @@
+from sqlalchemy.orm import Session
+
+import models
+import schemas
+
+
+def get_author_list(db: Session, skip: int = 0, limit: int = 100):
```
EdAlekseiev: Add type hint for return type here and for other functions

EdAlekseiev (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #151, opened 2023-11-07
Solution

grungezz:


yaroslav-mazurkevych (approved):

main.py
```diff
@@ -0,0 +1,53 @@
+from fastapi import FastAPI, Depends, HTTPException
+from sqlalchemy.orm import Session
+import crud
+import schemas
+from database import SessionLocal
+
```
Serhii-Leonenko, line 6: fix imports according to PEP

main.py
```diff
@@ -0,0 +1,53 @@
...
+def get_db() -> Session:
+    db = SessionLocal()
+    try:
+        yield db
+    finally:
+        db.close()
+
+
+@app.get("/authors", response_model=list[schemas.Author])
```
Serhii-Leonenko, line 18: don't forget to add` /` to the end of url

Serhii-Leonenko: the same in all other similar cases

main.py
Serhii-Leonenko: the same in all other similar cases

Serhii-Leonenko (changes requested):

PR #157, opened 2023-11-14
Develop

andy77andy:


crud.py
```diff
@@ -0,0 +1,69 @@
+from sqlalchemy.orm import Session
+
+from db import models, schemas
+
+
+def get_author_by_name(
+    db: Session,
+    name: str,
+):
```
ArsenPidhoretskyi: Provide return type annotations in all places

crud.py
```diff
@@ -0,0 +1,69 @@
...
+        queryset = (
+            db.query(models.DBBook)
+            .filter(models.DBBook.author_id == author_id)
+            .offset(page)
+            .limit(page_size)
+            .all()
+        )
+        return queryset
+
+    return db.query(models.DBBook).offset(page).limit(page_size).all()
```
ArsenPidhoretskyi: Here is a change only in one line(filter). Please no need to duplicate code

ArsenPidhoretskyi (changes requested):

crud.py
```diff
@@ -0,0 +1,67 @@
...
+
+from db import models, schemas
+from db.models import DBAuthor, DBBook
+
+
+def get_author_by_name(
+    db: Session,
+    name: str,
+) -> DBAuthor | None:
+    return db.query(models.DBAuthor).filter(models.DBAuthor.name == name).first()
```
ArsenPidhoretskyi: In this case in all places instead of models.DBAuthor you can use DBAuthor, the same with DBBook

crud.py
```diff
@@ -0,0 +1,67 @@
...
+    )
+    db.add(db_author)
+    db.commit()
+    db.refresh(db_author)
+
+    return db_author
+
+
+def get_all_books(
+    db: Session, author_id: int = None, page: int = 0, page_size: int = 20
```
ArsenPidhoretskyi:
```suggestion
    db: Session, author_id: Optional[int] = None, page: int = 0, page_size: int = 20
```
Because none is default parameter

crud.py
```diff
@@ -0,0 +1,67 @@
...
+) -> list[DBBook]:
+    queryset = db.query(models.DBBook)
+    if author_id:
+        queryset = (
+            db.query(models.DBBook)
```
ArsenPidhoretskyi:
```suggestion
        queryset = (
            queryset
```

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (approved):

PR #160, opened 2023-11-21
Solution

lotrik84:


crud.py
```diff
@@ -0,0 +1,41 @@
+from sqlalchemy.orm import Session
+
+from db import models
+import schemas
+
+
+def get_all_authors(db: Session):
```
ArsenPidhoretskyi: Provide for all functions return type annotations

ArsenPidhoretskyi (changes requested):

crud.py
```diff
@@ -0,0 +1,43 @@
+from typing import Any
+
+from sqlalchemy.orm import Session, Query
+
+from db import models
+import schemas
+
+
+def get_all_authors(db: Session) -> Query[Any]:
```
ArsenPidhoretskyi: Why do you use Any? It is bad practice!

ArsenPidhoretskyi (changes requested):

crud.py
lotrik84: I got it. I should use models class instead Any. Am I right?

vasylhnatiuk: Yes, you are right

vasylhnatiuk (changes requested):

main.py
```diff
@@ -0,0 +1,61 @@
...
+def read_root():
+    return {"Hello": "World"}
+
+
+@app.get("/authors/", response_model=Page[schemas.AuthorRead])
+def get_all_authors(db: Session = Depends(get_db)):
+    return paginate(crud.get_all_authors(db=db))
+
+
+@app.get("/authors/{author_id}")
```
vasylhnatiuk:
```suggestion
@app.get("/authors/{author_id}/")
```

alembic/README
```diff
@@ -0,0 +1 @@
+Generic single-database configuration.
```
vasylhnatiuk: add newline

vasylhnatiuk (changes requested):

crud.py
```diff
@@ -0,0 +1,41 @@
...
+
+from db import models
+import schemas
+
+
+def get_all_authors(db: Session) -> Query[models.DBAuthor]:
+    return db.query(models.DBAuthor).all()
+
+
+def get_author_by_id(db: Session, author_id: int) -> Query[models.DBAuthor]:
```
ArsenPidhoretskyi, line 11: It returns Optional[models.DBAuthor]

ArsenPidhoretskyi (approved):
Good job!

PR #171, opened 2023-12-22
Solution

D1st3f:


db/models.py
```diff
@@ -0,0 +1,24 @@
+from sqlalchemy import Column, Integer, String, ForeignKey, Date
+from sqlalchemy.orm import relationship
+
+from db.database import Base
+
+
+class Author(Base):
+    __tablename__ = "author"
```
wQuelS, line 8: we usually name tables in plural, for example "products" or "cars

db/models.py
```diff
@@ -0,0 +1,24 @@
...
+
+    id = Column(Integer, primary_key=True, index=True)
+    name = Column(String(255), nullable=False, unique=True)
+    bio = Column(String(1023), nullable=False)
+    books = relationship("Book", back_populates="author")
```
wQuelS, line 13: also, if you are using `SQLalchemy` v2.0+ - You'd better use new features like mapping.

main.py
```diff
@@ -0,0 +1,76 @@
+from fastapi import FastAPI, Depends, HTTPException
+from sqlalchemy.orm import Session
+
+import crud
+import schemas
```
wQuelS, line 5: do you really like those `.schemas` prefixes everywhere?

main.py
```diff
@@ -0,0 +1,76 @@
...
+
+app = FastAPI()
+
+
+def get_db() -> Session:
+    db = SessionLocal()
+    try:
+        yield db
+    finally:
+        db.close()
```
wQuelS, line 16: seems like this config should be placed somewhere else

main.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+@app.post("/authors/", response_model=schemas.Author)
+def create_author(author: schemas.AuthorCreate,
+                  db: Session = Depends(get_db)) -> schemas.Author:
+    db_author = crud.get_author_by_name(db=db, name=author.name)
+    if db_author is not None:
+        raise HTTPException(
+            status_code=400, detail="Author with this name already exists"
+        )
```
wQuelS, line 36: such logic should not be within the endpoint. Endpoints are for receiving requests and returning responses

schemas.py
```diff
@@ -0,0 +1,38 @@
...
+class AuthorCreate(AuthorBase):
+    pass
+
+
+class Author(AuthorBase):
+    id: int
+
+    class Config:
+        orm_mode = True
+
```
wQuelS, line 20: since pydantic v2.0 we'd better use `model_config = ConfigDict()` instead of `class Config`

image

crud.py
```diff
@@ -0,0 +1,53 @@
...
+
+import schemas
+from db.models import Author, Book
+
+
+def get_all_authors_with_pagination(db: Session,
+                                    skip: int = 0,
+                                    limit: int = 10) -> list[Author]:
+    authors = db.query(Author).offset(skip).limit(limit).all()
+    return authors
```
wQuelS, line 11: `query()` is gonna be deprecated. Use unified `select()`

crud.py
```diff
@@ -0,0 +1,53 @@
...
+def get_book_by_title(db: Session, title: str) -> Book:
+    return db.query(Book).filter(Book.title == title).first()
+
+
+def create_new_book_with_author(db: Session,
+                                book: schemas.BookCreate) -> Book:
+    db_book = Book(title=book.title,
+                   summary=book.summary,
+                   publication_date=book.publication_date,
+                   author_id=book.author_id)
```
wQuelS, line 49: this can be done in more pythonic way using new `pydantic` features. For example:

wQuelS (changes requested):

PR #172, opened 2023-12-25
solution

kostomeister:


mykolamateichuk (approved):

db/models.py
```diff
@@ -0,0 +1,26 @@
+from sqlalchemy import Column, Integer, String, ForeignKey, Text, Date
+from sqlalchemy.orm import relationship
+
+from db.database import Base
+
+
+class Book(Base):
+    __tablename__ = "book"
```
wQuelS, line 8: we usually name tables in plural like "products" or "cars"

db/models.py
```diff
@@ -0,0 +1,26 @@
...
+
+    id = Column(Integer, primary_key=True, index=True)
+    title = Column(String(255), nullable=False, unique=True)
+    summary = Column(Text, nullable=False, unique=True)
+    publication_date = Column(Date)
+    author_id = Column(Integer, ForeignKey("author.id"))
+
+    author = relationship("Author", back_populates="books")
```
wQuelS, line 16: also, if you are using `SQLalchemy` v2.0+ - u'd better use new features like mapping.

main.py
```diff
@@ -0,0 +1,85 @@
+from fastapi import FastAPI, Depends, HTTPException, Query
+from sqlalchemy.orm import Session
+
+import schemas
```
wQuelS, line 4: do you really like those `.schemas` prefixes everywhere?

main.py
```diff
@@ -0,0 +1,85 @@
...
+app = FastAPI()
+
+
+def get_db() -> Session:
+    db = SessionLocal()
+
+    try:
+        yield db
+    finally:
+        db.close()
```
wQuelS, line 17: this logic is better to keep somewhere else

main.py
```diff
@@ -0,0 +1,85 @@
...
+    author_id: int,
+    db: Session = Depends(get_db)
+):
+    author = crud.get_author(db=db, author_id=author_id)
+
+    if author is None:
+        raise HTTPException(
+            status_code=400,
+            detail="No such author!"
+        )
```
wQuelS, line 40: error-handling logic is not good to be kept within endpoints. Endpoints are for taking requests and returning responses. That's all

main.py
```diff
@@ -0,0 +1,85 @@
...
+    author: schemas.AuthorCreate,
+    db: Session = Depends(get_db)
+):
+    if crud.check_author_exists(db=db, author=author):
+        raise HTTPException(
+            status_code=400,
+            detail="Author with such name or bio already exists!"
+        )
+
+    return crud.create_author(db=db, author=author)
```
wQuelS, line 56: since you are not using `session` directly in your endpoints you have to pass it everytime to `crud` functions

schemas.py
```diff
@@ -0,0 +1,38 @@
...
+class BookCreate(BookBase):
+    author_id: int
+
+
+class BookList(BookBase):
+    id: int
+    author: AuthorList
+
+    class Config:
+        orm_mode = True
```
wQuelS, line 38: since `pydantic` v2.0 we'd better use `model_config = ConfigDict()` instead of `class Config`

image

crud.py
```diff
@@ -0,0 +1,73 @@
+from sqlalchemy import exists
+from sqlalchemy.orm import Session
+
+from db import models
+import schemas
+
+
+def get_all_authors(db: Session, skip: int = 0, limit: int = 10):
+    return db.query(models.Author).offset(skip).limit(limit).all()
```
wQuelS, line 9: `query()` is gonna be deprecated. Use unified `select()`

crud.py
```diff
@@ -0,0 +1,73 @@
...
+    return queryset.offset(skip).limit(limit).all()
+
+
+def create_book(db: Session, book: schemas.BookCreate):
+    db_book = models.Book(
+        title=book.title,
+        summary=book.summary,
+        publication_date=book.publication_date,
+        author_id=book.author_id
+    )
```
wQuelS, line 67: this can be done in a more pythonic way using new `pydantic` features. For example:

wQuelS (changes requested):

PR #173, opened 2023-12-25
Solution

HalynaPetrova:


crud.py
```diff
@@ -0,0 +1,80 @@
...
+        db: Session,
+        limit: int = 5,
+        offset: int = 0,
+
+) -> list[schemas.Author]:
+
+    return db.query(models.Author).limit(limit).offset(offset).all()
+
+
+def get_author_name(db: Session, name: str) -> models.Author:
```
ArsenPidhoretskyi: .first can return None also

ArsenPidhoretskyi (changes requested):

crud.py
HalynaPetrova: Sorry, but I didn't understand the question. Do you mean that it incorrectly checks when creating an author, whether such an author already exists? Or you are talking about pagination?

ArsenPidhoretskyi: No I mean that this method can return None also

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (approved):

PR #174, opened 2023-12-26
Solution

mykolamateichuk:


Familenko (changes requested):
Add migrations, otherwise looks like you did not test your own api

PR #176, opened 2023-12-26
Solution

vshvanska:


YuliiaSaliuta (approved):

crud.py
```diff
@@ -0,0 +1,53 @@
+from sqlalchemy.orm import Session
+from db import models
+from schemas import AuthorCreate, BookCreate
+
+
+def get_all_authors(db: Session, skip: int = 0, limit: int = 10):
+    return db.query(models.Author).offset(skip).limit(limit)
```
wQuelS: what version of SQLAlchemy are u using? If 2.0+ this query is depricated

db/models.py
```diff
@@ -0,0 +1,27 @@
...
+
+    books = relationship("Book", back_populates="author")
+
+
+class Book(Base):
+    __tablename__ = "books"
+
+    id = Column(Integer, primary_key=True, index=True)
+    title = Column(String(255))
+    summary = Column(String(511))
```
wQuelS: why 511?))

main.py
```diff
@@ -0,0 +1,64 @@
...
+    db = SessionLocal()
+    try:
+        yield db
+    finally:
+        db.close()
+
+
+@app.get("/")
+def read_root():
+    return {"Hello": "World"}
```
wQuelS: do you really need this endpoint?

main.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+@app.get("/authors/", response_model=list[schemas.Author])
+def read_authors(db: Session = Depends(get_db),
```
wQuelS, line 20: I know that this task was implemented according to the guide, but FYI: usually it is not good to carry session through all modules of your project. If you look carefully you'll mention that endpoints are not using it at all:)

This fix is up to you for now

main.py
```diff
@@ -0,0 +1,64 @@
+from fastapi import FastAPI, Depends, HTTPException, Query
+from sqlalchemy.orm import Session
+
+import crud
+import schemas
```
wQuelS: do you really like to add those schemas prefixes?:)

main.py
```diff
@@ -0,0 +1,64 @@
...
+def create_author(
+        author: schemas.AuthorCreate,
+        db: Session = Depends(get_db)
+):
+    db_author = crud.get_authors_by_name(db=db, name=author.name)
+    if db_author:
+        raise HTTPException(
+            status_code=400,
+            detail="Author with this name already exists"
+        )
```
wQuelS, line 41: I will not reject for this but FYI: such logic should never be at endpoints according to SOLID principles:)

schemas.py
```diff
@@ -0,0 +1,37 @@
...
+
+class AuthorCreate(AuthorBase):
+    pass
+
+
+class Author(AuthorBase):
+    id: int
+
+    class Config:
+        from_attributes = True
```
wQuelS: what version of pydantic are u using? If it is v2.0+ : such thing class config is deprecated and now we are using ConfigDict

P.s. `from_attributes` is correct, it is newer version of `orm_mode`, but `class Config` is an old style

schemas.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+class BookBase(BaseModel):
+    title: str
+    summary: str
+    publication_date: date
```
wQuelS, line 25: some of those are not required according to models. For example, `summary` is not required. So why is it strictly required here, in the serializer?

wQuelS (changes requested):

schemas.py
```diff
@@ -33,5 +33,5 @@ class Book(BookBase):
     id: int
     author: Author
 
-    class Config:
+    class ConfigDict:
```
wQuelS, line 36: it should be
`model_config = ConfigDict(from_attributes = True)`

https://docs.pydantic.dev/latest/migration/#changes-to-config

db/models.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+class Author(Base):
+    __tablename__ = "authors"
+
+    id = Column(Integer, primary_key=True, index=True)
+    name = Column(String(255), unique=True)
+    bio = Column(Text)
+
+    books = relationship("Book", back_populates="author")
```
wQuelS, line 14: also, if you are using SQLalchemy v2.0+ - u'd better use new features like mapping.

wQuelS (changes requested):

PR #180, opened 2023-12-26
Solution

YuliiaSaliuta:


Lyutillis (approved):

crud.py
```diff
@@ -0,0 +1,50 @@
...
+    return books
+
+
+def create_book(db: Session, book: schemas.BookCreate):
+    db_book = models.Book(
+        title=book.title,
+        summary=book.summary,
+        publication_date=book.publication_date,
+        author_id=book.author_id,
+    )
```
NEkropo1: consider usage:
```suggestion
    db_book = models.Book(
        **book.model_dump())
    )
```

main.py
```diff
@@ -0,0 +1,70 @@
...
+
+    try:
+        yield db
+    finally:
+        db.close()
+
+
+@app.get("/")
+def root() -> dict:
+    return {"message": "Hello World! There is Library API using FastAPI"}
```
NEkropo1: 'This is' or 'Here is' would be a little better, I think )

schemas.py
```diff
@@ -0,0 +1,36 @@
...
+
+class AuthorCreate(AuthorBase):
+    pass
+
+
+class Author(AuthorBase):
+    id: int
+
+    class Config:
+        orm_mode = True
```
NEkropo1:
```suggestion
        from_attributes = True
```
you can google, why I suggested it )

NEkropo1 (changes requested):

db/models.py
```diff
@@ -0,0 +1,24 @@
+from sqlalchemy import Column, Integer, String, Date, ForeignKey
+from sqlalchemy.orm import relationship
+
+from db.database import Base
+
+
+class Author(Base):
+    __tablename__ = "author"
```
wQuelS, line 8: We usually name tables in the plural. For example "products", "cars"

db/models.py
```diff
@@ -0,0 +1,24 @@
...
+
+    id = Column(Integer, primary_key=True, index=True)
+    name = Column(String(255), nullable=False, unique=True)
+    bio = Column(String(511), nullable=False)
+    books = relationship("Book", back_populates="author")
+
+
+class Book(Base):
+    __tablename__ = "book"
```
wQuelS, line 17: same with plural name

main.py
```diff
@@ -0,0 +1,70 @@
...
+
+    try:
+        yield db
+    finally:
+        db.close()
+
+
+@app.get("/")
+def root() -> dict:
+    return {"message": "Hello World! Here is Library API using FastAPI"}
```
wQuelS, line 23: do you need this endpoint?

main.py
```diff
@@ -0,0 +1,70 @@
+from fastapi import FastAPI, Depends, HTTPException, Query
+from sqlalchemy.orm import Session
+
+import schemas
```
wQuelS, line 4: do you really like that ".schemas" prefixes all the time?

main.py
```diff
@@ -0,0 +1,70 @@
...
+@app.post("/books/", response_model=schemas.Book)
+def create_book(
+    book: schemas.BookCreate,
+    db: Session = Depends(get_db),
+):
+    db_author = crud.get_author(db=db, author_id=book.author_id)
+    if not db_author:
+        raise HTTPException(
+            status_code=400, detail="Author with this id doesn't exist!"
+        )
```
wQuelS, line 69: usually such logic should be in other layers, like `service` or `application` layer. For now it is not mistake and fix is up to you

schemas.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+class AuthorCreate(AuthorBase):
+    pass
+
+
+class Author(AuthorBase):
+    id: int
+
+    class Config:
```
wQuelS, line 17: if you are using `pydantic` v.2.0+ than you should use `model_config = ConfigDict()` instead of `class Config`

db/models.py
```diff
@@ -0,0 +1,24 @@
...
+
+    id = Column(Integer, primary_key=True, index=True)
+    title = Column(String(255), nullable=False)
+    summary = Column(String(511), nullable=False)
+    publication_date = Column(DateTime)
+    author_id = Column(Integer, ForeignKey("author.id"))
+    author = relationship("Author", back_populates="books")
```
wQuelS, line 24: also, if you are using `SQLAlchemy` v2.0+ - you'd better use new features including `Mapped` and `mapping_column()`

crud.py
```diff
@@ -0,0 +1,47 @@
+from sqlalchemy.orm import Session
+
+import schemas
+from db import models
+
+
+def get_all_authors(db: Session, skip: int = 0, limit: int = 10):
+    return db.query(models.Author).offset(skip).limit(limit).all()
```
wQuelS, line 8: since new update `query()` is gonna be deprecated. Use general standard `select` now

wQuelS (changes requested):

PR #181, opened 2023-12-26
Solution

RabenkoYevhenii:


main.py
```diff
@@ -0,0 +1,60 @@
...
+    db = SessionLocal()
+    try:
+        yield db
+    finally:
+        db.close()
+
+
+@app.post("/authors/", response_model=schemas.Author)
+def create_author(author: schemas.AuthorCreate, db: Session = Depends(get_db)):
+    return crud.create_author(db=db, author=author)
```
kostomeister, line 35: Author's name is a unique field and when you create an author with an existing name you gonna get Internal server error, so you should handle it

kostomeister (changes requested):

library.db
wQuelS, line 1: never expose db files to the public! Remove it

main.py
```diff
@@ -0,0 +1,66 @@
...
+import schemas
+
+from database import SessionLocal
+from database import engine
+from fastapi import Depends
+from fastapi import FastAPI
+from fastapi import HTTPException
+from sqlalchemy.orm import Session
+
+models.Base.metadata.create_all(bind=engine)
```
wQuelS, line 14: why?

main.py
```diff
@@ -0,0 +1,66 @@
...
+
+app = FastAPI()
+
+
+def get_db():
+    db = SessionLocal()
+    try:
+        yield db
+    finally:
+        db.close()
```
wQuelS, line 24: seems like this can be in another place

main.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+@app.post("/authors/", response_model=schemas.Author)
+def create_author(author: schemas.AuthorCreate, db: Session = Depends(get_db)):
+    if crud.check_author_exists(db=db, author=author):
+        raise HTTPException(
+            status_code=400,
+            detail="Author with such name or bio already exists!"
+        )
```
wQuelS, line 33: error-handling logic is not good to be kept within endpoint function. Endpoints ( routers ) are for taking requests and returning responses. That's all

models.py
```diff
@@ -0,0 +1,29 @@
...
+class Book(Base):
+    __tablename__ = "books"
+
+    id = Column(Integer, primary_key=True, index=True)
+    title = Column(String)
+    summary = Column(String)
+    publication_date = Column(Date)
+
+    author_id = Column(Integer, ForeignKey("authors.id"))
+    author = relationship("Author", back_populates="books")
```
wQuelS, line 29: also, if you are using `SQLalchemy` v2.0+ - u'd better use new features like mapping.

schemas.py
```diff
@@ -0,0 +1,37 @@
...
+
+class AuthorCreate(AuthorBase):
+    pass
+
+
+class Author(AuthorBase):
+    id: int
+
+    class Config:
+        orm_mode = True
```
wQuelS, line 19: since `pydantic` v2.0 we'd better use `model_config = ConfigDict()` instead of `class Config`

image

crud.py
```diff
@@ -0,0 +1,48 @@
...
+def create_author(db: Session, author: Author):
+    db_author = Author(**author.dict())
+    db.add(db_author)
+    db.commit()
+    db.refresh(db_author)
+    return db_author
+
+
+def get_authors(db: Session, skip: int = 0, limit: int = 10):
+    return db.query(Author).offset(skip).limit(limit).all()
```
wQuelS, line 20: `query()` is gonna be deprecated. Use unified `select()`

crud.py
```diff
@@ -0,0 +1,48 @@
...
+import models
+import schemas
+from models import Author
+from models import Book
+from schemas import BookCreate
+from sqlalchemy.orm import Session
+
+
+def create_author(db: Session, author: Author):
+    db_author = Author(**author.dict())
```
wQuelS, line 12: it is gonna be deprecated. See screenshot below how this `dict()` method is changed now
image

wQuelS (changes requested):

PR #182, opened 2023-12-26
Solution

petrykivd:


SevKrok (approved):
The best 👍

crud.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+def get_author_by_name(db: Session, name: str):
+    return (
+        db.query(DBAuthor).filter(DBAuthor.name == name).first()
+    )
+
+
+def get_author_by_id(db: Session, author_id: int):
+    return db.query(DBAuthor).filter(DBAuthor.id == author_id).first()
```
wQuelS, line 18: what version of SQLAlchemy are u using? If 2.0+ this query is depricated

crud.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+def create_author(db: Session, author: AuthorCreate):
+    db_author = DBAuthor(
```
wQuelS, line 22: author.model_dump(), according to new pydantic v2.0

main.py
```diff
@@ -0,0 +1,89 @@
...
+
+    try:
+        yield db
+    finally:
+        db.close()
+
+
+@app.get("/")
+def root():
+    return {"message": "Hello World"}
```
wQuelS, line 25: do you really need this endpoint?

main.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+@app.get("/authors/", response_model=List[Author])
+def read_authors(
+        db: Session = Depends(get_db),
```
wQuelS, line 30: I know that this task was implemented according to the guide, but FYI: usually it is not good to carry session through all modules of your project. If you look carefully you'll mention that endpoints are not using it at all:)

This fix is up to you for now

main.py
```diff
@@ -0,0 +1,89 @@
...
+        author: AuthorCreate,
+        db: Session = Depends(get_db),
+):
+    db_author = crud.get_author_by_name(db=db, name=author.name)
+
+    if db_author:
+        raise HTTPException(
+            status_code=400,
+            detail="Such name for Author already exists"
+        )
```
wQuelS, line 58: I will not reject for this but FYI: such logic should never be at endpoints according to SOLID principles:)

main.py
```diff
@@ -0,0 +1,89 @@
...
+        limit=limit,
+    )
+
+
+@app.get("/books/{book_id}/", response_model=Book)
+def read_single_book(book_id: int, db: Session = Depends(get_db)):
+    db_book = crud.get_book(db=db, book_id=book_id)
+
+    if db_book is None:
+        raise HTTPException(status_code=404, detail="Book not found")
```
wQuelS, line 82: same as above

models.py
```diff
@@ -0,0 +1,33 @@
...
+    bio = Column(String(511), nullable=False)
+    books = relationship("DBBook", back_populates="author")
+
+
+class DBBook(Base):
+    __tablename__ = 'books'
+
+    id = Column(Integer, primary_key=True, index=True)
+    title = Column(String(255), nullable=False)
+    summary = Column(String(255), nullable=False)
```
wQuelS, line 29: was it made according to task description? Because non-nullable summary seems a bit weird

schemas.py
```diff
@@ -0,0 +1,37 @@
...
+
+class AuthorCreate(AuthorBase):
+    pass
+
+
+class Author(AuthorBase):
+    id: int
+
+    class Config:
+        orm_mode = True
```
wQuelS, line 19: what version of pydantic are u using? If it is v2.0+ : such thing class config is deprecated and now we are using ConfigDict

wQuelS (changes requested):

PR #183, opened 2023-12-26
solution

oleksandr-pupchenko:


Daniil-Pankieiev (approved):

crud.py
```diff
@@ -0,0 +1,50 @@
+from sqlalchemy.orm import Session
+
+from db.models import DBAuthor, DBBook
+from schemas import AuthorCreate, BookCreate
+
+
+def get_all_authors(db: Session, skip: int = 0, limit: int = 100):
+    return db.query(DBAuthor).offset(skip).limit(limit).all()
```
wQuelS, line 8: what version of SQLAlchemy are u using? If 2.0+ this query is depricated

crud.py
```diff
@@ -0,0 +1,50 @@
...
+        author_id=book.author_id,
+    )
+    db.add(db_book)
+    db.commit()
+    db.refresh(db_book)
+    return db_book
+
+
+def get_book_by_id(db: Session, author_id: int):
+    return db.query(DBBook).filter(DBBook.author_id == author_id).first()
```
wQuelS, line 50: why `.first()` ?

db/models.py
```diff
@@ -0,0 +1,26 @@
+from sqlalchemy import Column, Integer, String, ForeignKey, Date
+from sqlalchemy.orm import relationship
+
+from db.engine import Base
+
+
+class DBAuthor(Base):
+    __tablename__ = "author"
```
wQuelS, line 8: usually tables are named in plural

db/models.py
```diff
@@ -0,0 +1,26 @@
...
+
+    books = relationship("DBBook", back_populates="author")
+
+
+class DBBook(Base):
+    __tablename__ = "book"
+
+    id = Column(Integer, primary_key=True, index=True)
+    title = Column(String(255), nullable=False)
+    summary = Column(String(511), nullable=False)
```
wQuelS, line 22: was it according to task description? Because it is a bit weird that summary is non-nullable

main.py
```diff
@@ -0,0 +1,63 @@
...
+    db = SessionLocal()
+    try:
+        yield db
+    finally:
+        db.close()
+
+
+@app.get("/authors/", response_model=list[schemas.Author])
+def read_authors(
+    skip: int = 0, limit: int = 5, db: Session = Depends(get_db)
```
wQuelS, line 21: I know that this task was implemented according to the guide, but FYI: usually it is not good to carry session through all modules of your project. If you look carefully you'll mention that endpoints are not using it at all:)

This fix is up to you for now

main.py
```diff
@@ -0,0 +1,63 @@
+from fastapi import FastAPI, Depends, HTTPException
+from sqlalchemy.orm import Session
+
+import crud
+import schemas
```
wQuelS, line 5: do you reaally like those `schemas` prefixes?

main.py
```diff
@@ -0,0 +1,63 @@
...
+def create_author(author: schemas.AuthorCreate, db: Session = Depends(get_db)):
+    return crud.create_author(db=db, author=author)
+
+
+@app.get("/authors/{author_id}/", response_model=schemas.Author)
+def read_single_author(author_id: int, db: Session = Depends(get_db)):
+    db_author = crud.get_author_by_id(db=db, author_id=author_id)
+
+    if db_author is None:
+        raise HTTPException(status_code=404, detail="Author not found")
```
wQuelS, line 36: I will not reject for this but FYI: such logic should never be at endpoints according to SOLID principles:)

schemas.py
```diff
@@ -0,0 +1,38 @@
...
+
+class AuthorCreate(AuthorBase):
+    pass
+
+
+class Author(AuthorBase):
+    id: int
+
+    class Config:
+        from_attributes = True
```
wQuelS, line 19: as I see in reas I see in requirements.txt you use `pydantic v2.0+` and you've changed to new `from_attributes`. But you also forget that in pydantic v2.0+ `class Config` is deprecated, and no we use `ConfigDict`

wQuelS (changes requested):

PR #185, opened 2023-12-26
Solution

Daniil-Pankieiev:


Shatrovskyi (approved):

alembic/README
```diff
@@ -0,0 +1 @@
+Generic single-database configuration.
```
wQuelS, line 1: new line?

crud.py
```diff
@@ -0,0 +1,63 @@
...
+import models
+import schemas
+
+
+def get_all_authors(
+        db: Session,
+        skip: int = 0,
+        limit: int = 5
+):
+    return db.query(models.Author).offset(skip).limit(limit).all()
```
wQuelS, line 12: what version of SQLAlchemy are u using? If 2.0+ this query is depricated

database.py
```diff
@@ -0,0 +1,13 @@
...
+
+SQLALCHEMY_DATABASE_URL = "sqlite:///./library_service.db"
+
+engine = create_engine(
+    SQLALCHEMY_DATABASE_URL,
+    connect_args={"check_same_thread": False}
+)
+SessionLocal = sessionmaker(autocommit=False, autoflush=False, bind=engine)
+
+Base = declarative_base()
```
wQuelS, line 13: new line

main.py
```diff
@@ -0,0 +1,64 @@
+from fastapi import FastAPI, Depends, HTTPException
+from sqlalchemy.orm import Session
+
+import crud
+import schemas
```
wQuelS, line 5: do you really like to add those schemas prefixes?:)

main.py
```diff
@@ -0,0 +1,64 @@
...
+        yield db
+    finally:
+        db.close()
+
+
+@app.get("/authors/", response_model=list[schemas.Author])
+def read_authors(
+        skip: int = 0,
+        limit: int = 5,
+        db: Session = Depends(get_db)):
```
wQuelS, line 25: I know that this task was implemented according to the guide, but FYI: usually it is not good to carry session through all modules of your project. If you look carefully you'll mention that endpoints are not using it at all:)

This fix is up to you for now

main.py
```diff
@@ -0,0 +1,64 @@
...
+
+@app.post("/authors/", response_model=schemas.Author)
+def create_author(
+    author: schemas.AuthorCreate,
+    db: Session = Depends(get_db),
+):
+    db_author = crud.get_author_by_name(db=db, name=author.name)
+
+    if db_author:
+        raise HTTPException(status_code=400, detail="Such Author already exists")
```
wQuelS, line 37: I will not reject for this but FYI: such logic should never be at endpoints according to SOLID principles:)

and also it should be `status_code=status.HTTP_404_NOT_FOUND`

models.py
```diff
@@ -0,0 +1,25 @@
+from sqlalchemy import Column, Integer, String, ForeignKey, Date
+from sqlalchemy.orm import relationship
+
+from database import Base
+
+
+class Author(Base):
+    __tablename__ = "author"
```
wQuelS, line 8: we usually name tables in the plural. For example `products`

schemas.py
```diff
@@ -0,0 +1,39 @@
...
+
+class AuthorCreate(AuthorBase):
+    pass
+
+
+class Author(AuthorBase):
+    id: int
+
+    class Config:
+        orm_mode = True
```
wQuelS, line 19: You use pydantic v2.0. So `class Config` and `orm_mode` both are deprecated. Instead you should use `ConfigDict` and `from_attributes`
You may find more info in official docs

schemas.py
```diff
@@ -0,0 +1,39 @@
...
+class BookCreate(BookBase):
+    author_id: int
+
+
+class Book(BookBase):
+    id: int
+    author: Author
+
+    class Config:
+        orm_mode = True
```
wQuelS, line 39: new line

wQuelS (changes requested):

PR #187, opened 2023-12-27
Solution

Shatrovskyi:


YuliiaSaliuta (approved):

library.db
wQuelS, line 1: never expose `.db` files to the public! Remove it

main.py
```diff
@@ -0,0 +1,82 @@
+import crud
+import models
+from schemas import Author, Book, BookCreate, AuthorCreate
+from fastapi import FastAPI, Depends, HTTPException
+from sqlalchemy.orm import Session
+from database import engine, SessionLocal
+
+models.Base.metadata.create_all(bind=engine)
```
wQuelS, line 8: why?))

main.py
```diff
@@ -0,0 +1,82 @@
...
+
+app = FastAPI()
+
+
+def get_db() -> Session:
+    db = SessionLocal()
+    try:
+        yield db
+    finally:
+        db.close()
```
wQuelS, line 18: better to keep such logic in separate module

main.py
```diff
@@ -0,0 +1,82 @@
...
+def create_author(
+        author: AuthorCreate,
+        db: Session = Depends(get_db)
+) -> Author:
+    db_author = crud.get_author_by_name(db=db, name=author.name)
+
+    if db_author is not None:
+        raise HTTPException(
+            status_code=400, detail="Author already exists"
+        )
```
wQuelS, line 41: error handling logic is not that good to place in endpoints. Endpoints should be responsible for taking requests and returning responses. That's all

models.py
```diff
@@ -0,0 +1,24 @@
...
+
+class Book(Base):
+    __tablename__ = "books"
+
+    id = Column(Integer, primary_key=True, index=True)
+    title = Column(String(256))
+    summary = Column(String(1024))
+    publication_date = Column(Date)
+    author_id = Column(Integer, ForeignKey("authors.id"))
+    author = relationship(Author, back_populates="books")
```
wQuelS, line 24: also, if you are using `SQLalchemy` v2.0+ - u'd better use new features like mapping.

schemas.py
```diff
@@ -0,0 +1,37 @@
...
+
+class AuthorCreate(AuthorBase):
+    pass
+
+
+class Author(AuthorBase):
+    id: int
+
+    class Config:
+        from_attributes = True
```
wQuelS, line 18: since `pydantic` v2.0 we'd better use `model_config = ConfigDict()` instead of `class Config`

crud.py
```diff
@@ -0,0 +1,53 @@
+from sqlalchemy.orm import Session
+from models import Author, Book
+from schemas import AuthorCreate, BookCreate
+
+
+def get_author_by_id(db: Session, author_id: int):
+    return db.query(Author).filter(Author.id == author_id).first()
```
wQuelS, line 7: `query()` is gonna be deprecated. Use unified `select()`

crud.py
```diff
@@ -0,0 +1,53 @@
...
+    return queryset.offset(skip).limit(limit)
+
+
+def create_book(db: Session, book: BookCreate):
+    db_book = Book(
+        title=book.title,
+        summary=book.summary,
+        publication_date=book.publication_date,
+        author_id=book.author_id
+    )
```
wQuelS, line 49: this can be done in a more pythonic way using new `pydantic` features. For example:

wQuelS (changes requested):

PR #189, opened 2023-12-27
Develop

prochigor:


oleksandr-pupchenko (approved):

db/models.py
```diff
@@ -0,0 +1,27 @@
+from sqlalchemy import Column, Integer, String, Date, ForeignKey
+from sqlalchemy.orm import relationship
+
+from db.database import Base
+
+
+class DBAuthor(Base):
+    __tablename__ = "author"
```
wQuelS, line 8: we usually name tables in plural, for example "products" or "cars"

db/models.py
```diff
@@ -0,0 +1,27 @@
...
+
+    id = Column(Integer, primary_key=True, index=True)
+    name = Column(String(255), nullable=False, unique=True)
+    bio = Column(String(511), nullable=False)
+
+    books = relationship("DBBook", back_populates="author")
```
wQuelS, line 14: also, if you are using SQLalchemy v2.0+ - u'd better use new features like mapping.

db/models.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+class DBBook(Base):
+    __tablename__ = "book"
+
+    id = Column(Integer, primary_key=True, index=True)
+    title = Column(String(255), nullable=False)
+    summary = Column(String(511), nullable=False)
```
wQuelS, line 22: was it you idea to set `summary` as non-nullable or such requirements was in the task description?

main.py
```diff
@@ -0,0 +1,78 @@
+from fastapi import FastAPI, Depends, HTTPException
+from sqlalchemy.orm import Session
+
+import schemas
```
wQuelS, line 4: do you really like those `.schemas` prefixes everywhere?

main.py
```diff
@@ -0,0 +1,78 @@
...
+app = FastAPI()
+
+
+def get_db() -> Session:
+    db = SessionLocal()
+
+    try:
+        yield db
+    finally:
+        db.close()
```
wQuelS, line 17: seems like it is better to keep such thing somewhere else

main.py
```diff
@@ -0,0 +1,78 @@
...
+        author: schemas.AuthorCreate,
+        db: Session = Depends(get_db)
+):
+    db_author = crud.get_author_by_title(db=db, name=author.name)
+
+    if db_author:
+        raise HTTPException(
+            status_code=400,
+            detail="Such name for Author is already exist"
+        )
```
wQuelS, line 50: such logic should not be within `endpoint` ( router ). Endpoints are for receiving requests and returning responses.

For now it is not a mistake and I will not reject only for this issue, just saying

schemas.py
```diff
@@ -0,0 +1,37 @@
...
+class AuthorCreate(AuthorBase):
+    pass
+
+
+class Author(AuthorBase):
+    id: int
+
+    class Config:
+        orm_mode = True
+
```
wQuelS, line 20: since pydantic v2.0 we'd better use `model_config = ConfigDict()` instead of class Config.

image

crud.py
```diff
@@ -0,0 +1,62 @@
+from sqlalchemy.orm import Session
+
+from db import models
+import schemas
+
+
+def get_all_authors(db: Session, skip: int = 1, limit: int = 2):
+    return db.query(models.DBAuthor).offset(skip).limit(limit).all()
```
wQuelS, line 8: `query()` is gonna be deprecated. Use unified `select()`

crud.py
```diff
@@ -0,0 +1,62 @@
...
+    return db.query(models.DBBook).filter(models.DBBook.id == book_id).first()
+
+
+def create_book(db: Session, book: schemas.BookCreate):
+    db_book = models.DBBook(
+        title=book.title,
+        summary=book.summary,
+        publication_date=book.publication_date,
+        author_id=book.author_id,
+    )
```
wQuelS, line 58: this can be done in a more pythonic way using new `pydantic` features. For example:

wQuelS (changes requested):

PR #190, opened 2023-12-27
Solution

MiskoRuslan:


vshvanska (approved):

db/models.py
```diff
@@ -0,0 +1,24 @@
+from sqlalchemy import Column, Integer, String, Date, ForeignKey
+from sqlalchemy.orm import relationship
+
+from db.database import Base
+
+
+class Author(Base):
+    __tablename__ = "author"
```
wQuelS, line 8: we usually name tables in plural, for example `products`

db/models.py
```diff
@@ -0,0 +1,24 @@
...
+    bio = Column(String(511), nullable=True)
+    books = relationship("Book", back_populates="author")
+
+
+class Book(Base):
+    __tablename__ = "book"
+
+    id = Column(Integer, primary_key=True, index=True)
+    title = Column(String(255), unique=True, nullable=False)
+    summary = Column(String(511), nullable=True)
```
wQuelS, line 21: why 511?

main.py
```diff
@@ -0,0 +1,57 @@
+from fastapi import FastAPI, Depends, Query
+from sqlalchemy.orm import Session
+
+import crud
+import schemas
```
wQuelS, line 5: do you really like to use those `schemas` prefixes?

main.py
```diff
@@ -0,0 +1,57 @@
...
+        db.close()
+
+
+@app.get("/")
+def root():
+    return {"message": "hello world!"}
+
+
+@app.get("/authors/", response_model=list[schemas.AuthorList])
+def read_authors(db: Session = Depends(get_db)):
```
wQuelS, line 26: I know that this task was implemented according to the guide, but FYI: usually it is not good to carry session through all modules of your project. If you look carefully you'll mention that endpoints are not using it at all:)

This fix is up to you for now

main.py
```diff
@@ -0,0 +1,57 @@
...
+
+    try:
+        yield db
+    finally:
+        db.close()
+
+
+@app.get("/")
+def root():
+    return {"message": "hello world!"}
```
wQuelS, line 22: do you really need this endpoint?

schemas.py
```diff
@@ -0,0 +1,37 @@
...
+
+class AuthorCreate(AuthorBase):
+    pass
+
+
+class AuthorList(AuthorBase):
+    id: int
+
+    class Config:
+        orm_mode = True
```
wQuelS, line 19: what version of pydantic are u using? If it is v2.0+ : such thing class config is deprecated and now we are using ConfigDict

py-fastapi-library-management-api.db
wQuelS, line 1: all db files should be added to `.gitignore`. It is very unsafe to leak such files to public. Remove them, please

wQuelS (changes requested):

PR #192, opened 2023-12-28
Develop

sakhaline:


OlehShumov (approved):

crud.py
```diff
@@ -0,0 +1,47 @@
+from sqlalchemy.orm import Session
+
+import models
+import schemas
+
+
+def get_author_list(db: Session, skip: int = 0, limit: int = 5):
+    return db.query(models.Author).offset(skip).limit(limit).all()
```
wQuelS: what version of SQLAlchemy are u using? If 2.0+ this `query` is depricated
And even if you are using old version, try to use `select` and so on way of querying, since this casual `query` is gonna be deprecated.

main.py
```diff
@@ -0,0 +1,89 @@
+from fastapi import Depends, FastAPI, HTTPException, Query, Request, status
+from sqlalchemy.orm import Session
+from starlette.responses import JSONResponse
+
+import crud
+import schemas
```
wQuelS, line 5: do you really like to add those `schemas` prefixes?:)

main.py
```diff
@@ -0,0 +1,89 @@
...
+        status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
+        content={"field": error["loc"][-1], "message": error["msg"]},
+    )
+
+
+@app.get("/authors/", response_model=list[schemas.Author])
+def read_author_list(db: Session = Depends(get_db),
+                     skip: int = Query(0, ge=0),
+                     limit: int = Query(5)):
+    return crud.get_author_list(db=db, skip=skip, limit=limit)
```
wQuelS: I know that this task was implemented according to the guide, but FYI: usually it is not good to carry session through all modules of your project. If you look carefully you'll mention that endpoints are not using it at all:)

This fix is up to you for now

main.py
```diff
@@ -0,0 +1,89 @@
...
+
+@app.get("/authors/{author_id}/", response_model=schemas.Author)
+def read_author(db: Session = Depends(get_db), author_id: int | None = None):
+    author = crud.get_author_by_id(db=db, author_id=author_id)
+
+    if not author:
+        raise HTTPException(
+            status_code=status.HTTP_404_NOT_FOUND,
+            detail=f"Author with id {author_id} not found",
+        )
```
wQuelS: I will not reject for this but FYI: such logic should never be at endpoints according to SOLID principles:)

main.py
```diff
@@ -0,0 +1,89 @@
...
+
+    return author
+
+
+@app.post("/authors/",
+          response_model=schemas.Author,
+          status_code=status.HTTP_201_CREATED,
+          responses={422: {"model": schemas.ValidationErrorResponse,
+                           "description": "Validation Error"}})
+@unique_constraint_handler
```
wQuelS: don't like such an implementation, but again, not gonna reject this.
Some info you may take from something like this:
https://github.com/zhanymkanov/fastapi-best-practices

main.py
```diff
@@ -0,0 +1,89 @@
...
+                   author_id: int = None,
+                   skip: int = Query(0, ge=0),
+                   limit: int = Query(5)):
+
+    filtered_books = crud.filter_book_by_author_id(
+        author_id=author_id, db=db
+    )
+
+    if not list(filtered_books):
+        return crud.get_book_list(db=db, skip=skip, limit=limit)
```
wQuelS: There is too much redundant logic for a simple endpoint that should take a request and return a response.

models.py
```diff
@@ -0,0 +1,26 @@
...
+    bio = Column(String, nullable=False)
+
+    books = relationship("Book", back_populates="author")
+
+
+class Book(Base):
+    __tablename__ = "book"
+
+    id = Column(Integer, primary_key=True, index=True)
+    title = Column(String(255), nullable=False, unique=True, index=True)
```
wQuelS, line 21: just asking: what was the idea behind the decision to put an index on the `title` column?

models.py
```diff
@@ -0,0 +1,26 @@
...
+    summary = Column(String, nullable=False)
```
wQuelS: was it done according to the task description? Because it is a bit weird to make book's summary as non-nullable:)

schemas.py
```diff
@@ -0,0 +1,43 @@
...
+
+class AuthorCreate(AuthorBase):
+    pass
+
+
+class Author(AuthorBase):
+    id: int
+
+    class Config:
+        orm_mode = True
```
wQuelS: what version of pydantic are u using? If it is v2.0+ : such thing `class config` is deprecated and now we are using `ConfigDict`

validators.py
```diff
@@ -0,0 +1,19 @@
...
+        try:
+            return func(*args, **kwargs)
+        except sqlalchemy.exc.IntegrityError:
+            raise HTTPException(
+                status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
+                detail="Unique constraint violation. "
+                "The resource already exists.",
+            )
+
+    return wrapper
```
wQuelS: why have you decided to do it in such way?
The idea is good and validators should be made as separate modules, but your approach is a bit naive

wQuelS (changes requested):

validators.py
sakhaline: I wasn't sure about this decision. 
Is it a bad idea to use decorators to catch errors because if there are more errors, then there will be too many decorators on top of functions?)) Is it better to use separate functions and use them inside endpoints?

models.py
sakhaline: That's disgusting...

models.py
sakhaline: Search optimization. 
We most often search for a book by title

main.py
sakhaline: yep)
well actually I don't like massive imports

main.py
sakhaline: thanks

validators.py
sakhaline: I wasn't sure about this decision...

LiudmylaKulinchenko (approved):
Well done

PR #193, opened 2023-12-28
Solution

OlehShumov:


db/models.py
```diff
@@ -0,0 +1,24 @@
+from sqlalchemy import Column, Integer, String, ForeignKey, Date
+from sqlalchemy.orm import relationship
+
+from .engine import Base
+
+
+class DBAuthor(Base):
+    __tablename__ = "author"
```
wQuelS: usually, we name tables in plural since it is a representation of "authors" in this case

db/models.py
```diff
@@ -0,0 +1,24 @@
...
+
+    id = Column(Integer, primary_key=True, index=True)
+    name = Column(String(63), nullable=False, unique=True)
+    bio = Column(String(255), nullable=False)
+    books = relationship("DBBook", back_populates="author")
+
+
+class DBBook(Base):
+    __tablename__ = "book"
```
wQuelS: same

db/models.py
```diff
@@ -0,0 +1,24 @@
...
+
+    id = Column(Integer, primary_key=True, index=True)
+    title = Column(String(63), nullable=False)
+    summary = Column(String(255), nullable=False)
```
wQuelS: was it your idea to make `summary` as non-nullable? Because it sounds weird a bit

main.py
```diff
@@ -0,0 +1,59 @@
+from fastapi import FastAPI
+from db.engine import session
+from fastapi import Depends
+from fastapi.responses import RedirectResponse
+from fastapi import HTTPException
+from starlette.status import HTTP_404_NOT_FOUND
+
+import schemas
+import crud
```
wQuelS: do you really like to add `schemas` or `crud` prefixes everywhere?

main.py
```diff
@@ -0,0 +1,59 @@
...
+    db = session()
+    try:
+        yield db
+    finally:
+        db.close()
+
+
+@app.get("/", include_in_schema=False)
+def get_docs():
+    return RedirectResponse(url="/docs")
```
wQuelS, line 29: huh?))

main.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+@app.get("/authors/{author_id}/", response_model=schemas.Author)
+def get_author_by_id(author_id, db=Depends(get_db)):
+    author = crud.get_author_by_id(db=db, author_id=author_id)
+    if not author:
+        raise HTTPException(
+            status_code=HTTP_404_NOT_FOUND,
+            detail="Author with this id does not exist",
+        )
```
wQuelS: I will not reject for this but FYI: such logic should never be at endpoints according to SOLID principles:)

main.py
```diff
@@ -0,0 +1,59 @@
...
+        db.close()
+
+
+@app.get("/", include_in_schema=False)
+def get_docs():
+    return RedirectResponse(url="/docs")
+
+
+@app.get("/authors/", response_model=list[schemas.Author])
+def get_authors(db=Depends(get_db)):
```
wQuelS: 1) `db: Session = Depends(get_db)`   to be clear
2) I know that this task was implemented according to the guide, but FYI: usually it is not good to carry session through all modules of your project. If you look carefully you'll mention that endpoints are not using it at all:)

Fix `2)` is up to you, will not reject for this

main.py
```diff
@@ -0,0 +1,59 @@
+from fastapi import FastAPI
+from db.engine import session
+from fastapi import Depends
+from fastapi.responses import RedirectResponse
+from fastapi import HTTPException
+from starlette.status import HTTP_404_NOT_FOUND
```
wQuelS: imports order?

schemas.py
```diff
@@ -0,0 +1,36 @@
...
+
+class AuthorCreate(AuthorBase):
+    pass
+
+
+class Author(AuthorBase):
+    id: int
+
+    class Config:
+        orm_mode = True
```
wQuelS: what version of pydantic are u using? If it is v2.0+ than such thing is depricated and now we are using `ConfigDict`

wQuelS (changes requested):

db/models.py
```diff
@@ -0,0 +1,24 @@
...
+from .engine import Base
+
+
+class DBAuthor(Base):
+    __tablename__ = "authors"
+
+    id = Column(Integer, primary_key=True, index=True)
+    name = Column(String(63), nullable=False, unique=True)
+    bio = Column(String(255), nullable=False)
+    books = relationship("DBBook", back_populates="author")
```
wQuelS, line 13: why not using `SQLAlchemy` v2.0 + `Mapped` feature?

schemas.py
```diff
@@ -0,0 +1,36 @@
...
+
+class AuthorCreate(AuthorBase):
+    pass
+
+
+class Author(AuthorBase):
+    id: int
+
+    class ConfigDict:
+        orm_mode = True
```
wQuelS, line 18: this attribute now callse `from_attributes` and it should be like this:
`model_config = ConfigDict(from_attributes=True)`

crud.py
```diff
@@ -0,0 +1,47 @@
...
+from sqlalchemy.orm import Session
+from db import models
+from schemas import AuthorCreate, BookCreate
+
+
+def get_author_by_id(db: Session, author_id):
+    author = db.query(models.DBAuthor).filter(models.DBAuthor.id == author_id).first()
+    if not author:
+        raise HTTPException(status_code=404, detail="Author not found")
+    return author
```
wQuelS, line 11: `query()` is gonna be deprecated. Use unified `select()`

crud.py
```diff
@@ -0,0 +1,47 @@
...
+    return db.query(models.DBBook).filter(models.DBBook.author_id == author_id).all()
+
+
+def create_book(db: Session, book: BookCreate):
+    db_book = models.DBBook(
+        title=book.title,
+        summary=book.summary,
+        publication_date=book.publication_date,
+        author_id=book.author_id
+    )
```
wQuelS, line 43: this can be done in a more pythonic way using new `pydantic` features. For example:

wQuelS (changes requested):

PR #194, opened 2023-12-29
Develop

ZaikaDaria:


alembic/README
```diff
@@ -0,0 +1 @@
+Generic single-database configuration.
```
vasylhnatiuk: add newline

vasylhnatiuk (changes requested):
provide return annotation for all of functions

vasylhnatiuk (approved):

PR #195, opened 2024-01-05
Create first website with fast api

Sel-Fisher:


SevKrok (approved):

db/models.py
```diff
@@ -0,0 +1,22 @@
+from sqlalchemy import Column, Integer, String, Date, ForeignKey
+from sqlalchemy.orm import relationship
+
+from db.database import Base
+
+
+class DBAuthor(Base):
+    __tablename__ = "author"
```
wQuelS, line 8: we usually name our tables in plural, for example "products" or "cars"

db/models.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+class DBBook(Base):
+    __tablename__ = "book"
+    id = Column(Integer, primary_key=True, index=True)
+    title = Column(String(255), nullable=False)
+    summary = Column(String(511), nullable=False)
+    publication_date = Column(Date, nullable=False)
+    author_id = Column(Integer, ForeignKey("author.id"))
+    author = relationship("DBAuthor", back_populates="books")
```
wQuelS, line 22: also, if you are using SQLalchemy v2.0+ - u'd better use new features like mapping.

db/models.py
```diff
@@ -0,0 +1,22 @@
...
+from sqlalchemy.orm import relationship
+
+from db.database import Base
+
+
+class DBAuthor(Base):
+    __tablename__ = "author"
+    id = Column(Integer, primary_key=True, index=True)
+    name = Column(String(255), nullable=False, unique=True)
+    bio = Column(String(511), nullable=False)
```
wQuelS, line 11: is it your decision to make `bio` non-nullable or it is requirement from task description?

main.py
```diff
@@ -0,0 +1,81 @@
...
+
+app = FastAPI()
+
+
+def get_db() -> Session:
+    db = SessionLocal()
+    try:
+        yield db
+    finally:
+        db.close()
```
wQuelS: seems like this thing should be moved to separate module?

main.py
```diff
@@ -0,0 +1,81 @@
+from datetime import date
+
+from fastapi import FastAPI, Depends, HTTPException, Query
+from sqlalchemy.orm import Session
+
+import crud
+import schemas
```
wQuelS: do you really like those `.schemas` prefixes everywhere?

main.py
```diff
@@ -0,0 +1,81 @@
...
+    db = SessionLocal()
+    try:
+        yield db
+    finally:
+        db.close()
+
+
+@app.get("/")
+def root() -> dict:
+    return {"message": "Hello World!"}
```
wQuelS: do you really need this endpoint?

main.py
```diff
@@ -0,0 +1,81 @@
...
+    return crud.get_all_authors(db=db, book=book, skip=skip, limit=limit)
+
+
+@app.post("/authors/", response_model=schemas.Author)
+def create_author(author: schemas.AuthorCreate, db: Session = Depends(get_db)):
+    db_author = crud.get_author_by_name(db=db, name=author.name)
+    if db_author:
+        raise HTTPException(
+            status_code=400, detail="Author with this name already exists"
+        )
```
wQuelS, line 29: seems like error handling logic is not that good to place in endpoints. Endpoints should be responsible for taking requests and returning responses

main.py
```diff
@@ -0,0 +1,81 @@
...
+    return crud.create_author(db=db, author=author)
+
+
+@app.get("/authors/{author_id}/", response_model=schemas.Author)
+def get_single_author(author_id: int, db: Session = Depends(get_db)):
+    db_author = crud.get_author_by_id(db=db, author_id=author_id)
```
wQuelS, line 35: since session is not being used directly at endpoints you have to pass it all the time to `crud` functions

schemas.py
```diff
@@ -0,0 +1,37 @@
...
+
+class BookCreate(BookBase):
+    author_id: int
+
+
+class Book(BookBase):
+    id: int
+
+    class Config:
+        orm_mode = True
```
wQuelS: since pydantic v2.0 we'd better use model_config = ConfigDict() instead of class Config.

image

crud.py
```diff
@@ -0,0 +1,80 @@
...
+from sqlalchemy.orm import Session
+
+from db import models
+import schemas
+
+
+def get_all_authors(
+    db: Session, limit: int = 10, skip: int = 0, book: str | None = None
+):
+    queryset = db.query(models.DBAuthor)
```
wQuelS: `query()` is gonna be deprecated. Use unified `select()`

crud.py
```diff
@@ -0,0 +1,80 @@
...
+    return db.query(models.DBBook).filter(models.DBBook.id == book_id).first()
+
+
+def create_book(db: Session, book: schemas.BookCreate):
+    db_book = models.DBBook(
+        title=book.title,
+        summary=book.summary,
+        publication_date=book.publication_date,
+        author_id=book.author_id,
+    )
```
wQuelS: this can be done in more pytonic way using new pydantic features. For example:

wQuelS (changes requested):

schemas.py
LiudmylaKulinchenko, line 1: no blank line

LiudmylaKulinchenko (changes requested):
There are some comments from Oleksandr that you didn't fix and didn't answer

PR #199, opened 2024-01-18
Solution

amirammina:


crud.py
```diff
@@ -0,0 +1,49 @@
+from sqlalchemy.orm import Session
+
+import schemas
+from db import models
+
+
+def get_author_list(db: Session):
```
ArsenPidhoretskyi: provide return type annotations

ArsenPidhoretskyi (changes requested):

crud.py
```diff
@@ -0,0 +1,53 @@
+from typing import Type
+
+from sqlalchemy.orm import Session
+
+import schemas
+from db import models
+
+
+def get_author_list(db: Session) -> list[Type[models.DBAuthor]]:
```
ArsenPidhoretskyi: Really it returns Type?

ArsenPidhoretskyi: Fix in all places

ArsenPidhoretskyi (changes requested):

crud.py
amirammina: I was surprised as much as you, but it was PyCharm suggestion :)

crud.py
```diff
@@ -0,0 +1,50 @@
+from fastapi import Query
+from sqlalchemy.orm import Session
+
+import schemas
+from db.models import DBAuthor, DBBook
+
+
+def get_author_list(db: Session) -> Query:
```
ArsenPidhoretskyi: Is .all return query?

crud.py
```diff
@@ -0,0 +1,50 @@
...
+    return db.query(DBAuthor).all()
+
+
+def get_author(db: Session, author_id: int) -> Query:
```
ArsenPidhoretskyi: Is .first return query?

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #200, opened 2024-01-19
Develop

roma-mechanic:


sql_app/crud.py
```diff
@@ -0,0 +1,52 @@
+from sqlalchemy.orm import Session
+
+from sql_app import models, schemas
+
+
+def create_author(db: Session, author: schemas.AuthorCreate) -> schemas.AuthorRetrieve:
```
ArsenPidhoretskyi: Really? As for me, it returns models.Author

ArsenPidhoretskyi: Wrong return type annotations in all places

sql_app/crud.py
```diff
@@ -0,0 +1,52 @@
...
+    db.commit()
+    db.refresh(db_author)
+    return db_author
+
+
+def get_all_authors(db: Session, skip: int = 0, limit: int = 100) -> list[schemas.AuthorRetrieve]:
+    return db.query(models.Author).offset(skip).limit(limit).all()
+
+
+def get_author_by_id(db: Session, author_id: int) -> schemas.AuthorRetrieve:
```
ArsenPidhoretskyi: btw, it can return None

ArsenPidhoretskyi: read documentation what .first can return

ArsenPidhoretskyi (changes requested):
sql_app is so bad name for application

sql_app/crud.py
roma-mechanic: .first() return the first result of this Query or None if the result doesn’t contain any row.

NikOneZ1 (approved):

PR #221, opened 2024-03-09
Solution

OlehOryshchuk:


vasylhnatiuk (changes requested):
provide return annotation for all functions

EdAlekseiev (approved):

PR #231, opened 2024-04-03
Pochtalon

pochtalon:


lib_app/main.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+@app.get("/")
+def read_root():
+    return {"Hello": "World"}
+
+
+@app.get("/items/{item_id}")
+def read_item(item_id: int, q: Union[str, None] = None):
+    return {"item_id": item_id, "q": q}
```
YuriiSmolii: this code should be removed.

YuriiSmolii (changes requested):
Add pagination (skip, limit) for authors and books endpoints.

NazarSheremet (approved):
Good job!

PR #233, opened 2024-04-09
Solution

EugeneProtsenko:


crud.py
```diff
@@ -0,0 +1,52 @@
+from sqlalchemy.orm import Session
+
+from db import models
+import schemas
+
+
+def get_all_authors(db: Session, skip: int = 0, limit: int = 10):
```
ArsenPidhoretskyi: Specify return type annotations in all functions

ArsenPidhoretskyi (changes requested):

EdAlekseiev (approved):

PR #235, opened 2024-04-12
Solution

aLEKS-e3:


crud.py
```diff
@@ -0,0 +1,28 @@
...
+    return db.query(Book).all()
+
+
+def create_book(db: Session, book: BookCreate):
+    db_book = Book(**book.dict())
+    db.add(db_book)
+    db.commit()
+    db.refresh(db_book)
+
+    return db_book
```
MNDonut: let's get db as the first parameter in each helper method to be consistent

main.py
```diff
@@ -0,0 +1,97 @@
...
+from fastapi_pagination import paginate, add_pagination, Page
+from sqlalchemy.orm import Session
+
+import crud
+import pagination
+import schemas
+from database import SessionLocal
+from models import Book
+
+app = FastAPI()
```
MNDonut: 2 blank lines between the last import and the first line of code

main.py
```diff
@@ -0,0 +1,97 @@
...
+):
+    return paginate(crud.get_books(author_id, db))
+
+
+def get_book_by_title(title: str, db: Session = Depends(get_db)):
+    return (
+        db.query(Book)
+        .filter(Book.title == title)
+        .first()
+    )
```
MNDonut: let's move it to helper methods file and make db as the first parameter

main.py
```diff
@@ -0,0 +1,97 @@
...
+    db = SessionLocal()
+    try:
+        yield db
+    finally:
+        db.close()
+
+
+@app.get("/")
+def read_root():
+    return {"Hello": "World"}
```
MNDonut: you can remove this endpoint

main.py
```diff
@@ -0,0 +1,97 @@
...
+
+        if get_book_by_title(book.title, db):
+            raise HTTPException(
+                status_code=400,
+                detail="Book with that title already exists",
+            )
+
+        book_data = book.dict()
+        book_data["author_id"] = author_id
+        book = schemas.BookCreate(**book_data)
```
MNDonut: or you can just pass author_id=author_id directly in BookCreate

MNDonut (changes requested):

main.py
```diff
@@ -19,36 +19,23 @@ def get_db() -> Session:
...
-    return paginate(crud.get_books(author_id, db))
-
-
-def get_book_by_title(title: str, db: Session = Depends(get_db)):
-    return (
-        db.query(Book)
-        .filter(Book.title == title)
-        .first()
-    )
+    return paginate(crud.get_books(author_id=author_id, db=db))
```
MNDonut, line 27: no need to pass parameters' name explicitly, just reorder them so that it is in the same order that in function

main.py
```diff
@@ -19,36 +19,23 @@ def get_db() -> Session:
...
 def create_book(book: schemas.BookCreate, db: Session = Depends(get_db)):
-    if get_book_by_title(book.title, db):
+    if crud.get_book_by_title(db, book.title):
         raise HTTPException(
             status_code=400,
             detail="Book with that title already exists",
         )
 
-    return crud.create_book(db, book)
+    return crud.create_book(db=db, book=book)
```
MNDonut, line 38: no need to pass parameters' name explicitly, just reorder them so that it is in the same order that in function

main.py
```diff
@@ -58,7 +45,7 @@ def get_authors_list(db: Session = Depends(get_db)):
 
 @app.get("/authors/{author_id}/", response_model=schemas.AuthorReadDetail)
 def get_author_detail(author_id: int, db: Session = Depends(get_db)):
-    if author := crud.get_author_by_id(db, author_id):
+    if author := crud.get_author_by_id(db=db, author_id=author_id):
```
MNDonut, line 48: no need to pass parameters' name explicitly, just reorder them so that it is in the same order that in function

MNDonut (approved):

PR #241, opened 2024-04-15
Solution

kabust:
image


crud.py
```diff
@@ -0,0 +1,58 @@
+from sqlalchemy.orm import Session
+
+import models, schemas
+
+
+def get_author(db: Session, author_id: int):
```
Y-Havryliv:
```suggestion
def get_author_by_id(db: Session, author_id: int):
```

library.db
Y-Havryliv, line 1: delete the db from PR

Y-Havryliv (changes requested):

Y-Havryliv (approved):
nice!

PR #258, opened 2024-04-18
solution

Serejka21:


app/models.py
```diff
@@ -0,0 +1,24 @@
...
+from sqlalchemy.orm import relationship
+
+from app.database import Base
+
+
+class Book(Base):
+    __tablename__ = "books"
+
+    id = Column(Integer, primary_key=True, index=True)
+    tittle = Column(String(63), nullable=False)
```
MNDonut: misspelled field name + it shouldn't be nullable per requirements

app/models.py
```diff
@@ -0,0 +1,24 @@
...
+    summary = Column(String(255), )
+    publication_date = Column(DateTime, )
+    author_id = Column(Integer, ForeignKey("authors.id"))
+
+
+class Author(Base):
+    __tablename__ = "authors"
+
+    id = Column(Integer, primary_key=True, index=True)
+    name = Column(String(63), nullable=False, unique=True)
```
MNDonut, line 22: it shouldn't be nullable

main.py
```diff
@@ -0,0 +1,76 @@
...
+
+@app.get("/authors/{author_id}/", response_model=schemas.Author)
+def read_single_author(author_id: int, db: Session = Depends(get_db)):
+    if db_author := crud.get_author(db=db, author_id=author_id):
+        return db_author
+
+    raise HTTPException(
+        status_code=404,
+        detail="Author not found"
+    )
```
MNDonut, line 53: This functionality is not implemented "Filter books by author ID."

MNDonut (changes requested):

main.py
Serejka21: image

MNDonut (approved):

PR #299, opened 2024-06-26
solution

SerhiiMar:


.gitignore
```diff
@@ -128,3 +128,6 @@ dmypy.json
 # Pyre type checker
 .pyre/
 .idea/
+
+# project database
+library.db
```
LiudmylaKulinchenko: You should add all files with db extention to the gitignore file here

LiudmylaKulinchenko (changes requested):

fsocie7y (approved):
good now

PR #301, opened 2024-06-27
Solution

vkleshko:


main.py
```diff
@@ -0,0 +1,64 @@
...
+def get_db() -> Session:
+    db = SessionLocal()
+
+    try:
+        yield db
+    finally:
+        db.close()
+
+
+@app.get("/authors", response_model=List[schemas.Author])
```
vasylhnatiuk:
```suggestion
@app.get("/authors/", response_model=List[schemas.Author])
```

vasylhnatiuk: all paths should be with / in the end

vasylhnatiuk (changes requested):
provide return annotations for each function

vasylhnatiuk (approved):

PR #335, opened 2024-08-27
sol1

IvanVahanov:


db/main.py
```diff
@@ -0,0 +1,52 @@
...
+
+@app.post("/authors/", response_model=schemas.Author)
+def create_author(author: schemas.AuthorCreate, db: Session = Depends(get_db)):
+    return crud.create_author(db=db, author=author)
+
+
+@app.get("/authors/", response_model=list[schemas.Author])
+def read_authors(skip: int = 0, limit: int = 10, db: Session = Depends(get_db)):
+    authors = crud.get_authors(db, skip=skip, limit=limit)
+    return authors
```
DevSpaciX: Provide return type annotations in all places

DevSpaciX (changes requested):
Remove .db file from PR

PR #342, opened 2024-08-30
solution

Yuruch:


alembic/env.py
```diff
@@ -0,0 +1,80 @@
...
+# Interpret the config file for Python logging.
+# This line sets up loggers basically.
+if config.config_file_name is not None:
+    fileConfig(config.config_file_name)
+
+# add your model's MetaData object here
+# for 'autogenerate' support
+# from myapp import mymodel
+# target_metadata = mymodel.Base.metadata
+target_metadata = [Base.metadata]
```
maxgerman: shouldn't be a list

crud.py
```diff
@@ -0,0 +1,51 @@
+from fastapi import Query
+from sqlalchemy.orm import Session
+
+import models
+import schemas
+
+
+def get_authors(db: Session, skip: int = 0, limit: int = 100) -> Query:
+    query = db.query(models.Author).offset(skip).limit(limit)
```
maxgerman: query is outdated style - check out the current docs (>2.0); `select` is used for both ORM and Core nowadays

crud.py
```diff
@@ -0,0 +1,51 @@
+from fastapi import Query
+from sqlalchemy.orm import Session
+
+import models
+import schemas
+
+
+def get_authors(db: Session, skip: int = 0, limit: int = 100) -> Query:
```
maxgerman: the return type is incorrect, we return the objects not a query

crud.py
```diff
@@ -0,0 +1,51 @@
...
+def create_author(db: Session, author: schemas.AuthorCreate) -> models.Author:
+    db_author = models.Author(name=author.name, bio=author.bio)
+    db.add(db_author)
+    db.commit()
+    db.refresh(db_author)
+
+    return db_author
+
+
+def get_author_by_name(db: Session, name: str) -> models.Author:
```
maxgerman: the return type should account for the case when there's no such author

crud.py
```diff
@@ -0,0 +1,51 @@
...
+    return db.query(models.Author).filter(models.Author.name == name).first()
+
+
+def get_author_by_id(db: Session, author_id: int) -> models.Author:
```
maxgerman: same as above

main.py
```diff
@@ -0,0 +1,56 @@
...
+@app.get("/authors/{author_id}", response_model=schemas.Author)
+def read_author(author_id: int, db: Session = Depends(get_db)) -> models.Author:
+    db_author = crud.get_author_by_id(db=db, author_id=author_id)
+    if db_author is None:
+        raise HTTPException(status_code=404, detail="Author not found")
+    return db_author
+
+
+@app.post("/authors/", response_model=schemas.Author)
+def create_author(author: schemas.AuthorCreate, db: Session = Depends(get_db)) -> models.Author:
```
maxgerman: http status code for creation?

schemas.py
```diff
@@ -0,0 +1,38 @@
...
+class BookCreate(BookBase):
+    pass
+
+
+class Book(BookBase):
+    id: int
+    author_id: int
+
+    class Config:
+        orm_mode = True
```
maxgerman: this is the old style (pre-2.0) - see the docs - there's a migration guide in pydantic 2.0 docs specifically for such things

maxgerman (changes requested):
Good!
See some comments.
You shouldn't commit directly to your master/main branch - use the dev branch.
Use async style where possible (requests, db) - it's a good match for fastapi.

Yuruch:
I guess async is unnecessary for this task because the videos did not mention that. But if u want I can change it

crud.py
```diff
@@ -1,16 +1,24 @@
...
 from sqlalchemy.orm import Session
 
 import models
 import schemas
 
 
-def get_authors(db: Session, skip: int = 0, limit: int = 100) -> Query:
-    query = db.query(models.Author).offset(skip).limit(limit)
-    return query.all()
+def get_authors(
```
maxgerman: use async session and async style db-queries - async support is fastapi's strength, let's use it

crud.py
```diff
@@ -1,16 +1,24 @@
...
+        db: Session,
+        skip: int = 0,
```
maxgerman, line 10: these pagination parameters are the same for all typical list endpoints - you could refactor them to a single fastapi dependency and avoid repetition

main.py
```diff
@@ -18,21 +18,43 @@ def get_db() -> Session:
...
+) -> list:
     return crud.get_authors(db=db, skip=skip, limit=limit)
 
 
-@app.get("/authors/{author_id}", response_model=schemas.Author)
-def read_author(author_id: int, db: Session = Depends(get_db)) -> models.Author:
+@app.get(
+    "/authors/{author_id}",
+    response_model=schemas.Author,
+    status_code=status.HTTP_200_OK
```
maxgerman: this is default status code, not necessary to explicitly state it (unlike non-200 ones)

maxgerman (changes requested):
Good overall, let's add async session and use it

Arsen-hrynevych (changes requested):
Please review your code and ensure that all operations support asynchronous execution. For instance, the Alembic configuration is not asynchronous.

Arsen-hrynevych (approved):
Thanks

PR #346, opened 2024-09-04
solution

DimaYankovskiy:


alembic/env.py
```diff
@@ -0,0 +1,74 @@
...
+# Interpret the config file for Python logging.
+# This line sets up loggers basically.
+if config.config_file_name is not None:
+    fileConfig(config.config_file_name)
+
+# add your model's MetaData object here
+# for 'autogenerate' support
+# from myapp import mymodel
+# target_metadata = mymodel.Base.metadata
+target_metadata = [Base.metadata]
```
maxgerman, line 24: doesn't need to be a list

crud.py
```diff
@@ -0,0 +1,51 @@
+from fastapi import Query
+from sqlalchemy.orm import Session
+
+import models
+import schemas
+
+
+def get_all_authors(db: Session, skip: int = 0, limit: int = 100) -> Query:
```
maxgerman, line 8: - use async sytle (plays nice with fastapi and async libs available)
- these pagination parameters are the same for all typical list endpoints - you could refactor them to a single fastapi dependency and avoid repetition
- query is outdated style - check out the current docs (>2.0) (use `select`)

database.py
```diff
@@ -0,0 +1,13 @@
...
+from sqlalchemy.ext.declarative import declarative_base
+from sqlalchemy.orm import sessionmaker
+
+SQLALCHEMY_DATABASE_URL = "sqlite:///./library.db"
+
+engine = create_engine(
+    SQLALCHEMY_DATABASE_URL, connect_args={"check_same_thread": False}
+)
+
+SessionLocal = sessionmaker(autocommit=False, autoflush=False, bind=engine)
```
maxgerman, line 11: use async session

schemas.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+class AuthorCreate(AuthorBase):
+    pass
+
+
+class Author(AuthorBase):
+    id: int
+
+    class Config:
```
maxgerman, line 18: this is the old style (pre-2.0) - see the pydantic docs or migration guide

maxgerman (changes requested):
Good, see some comments

PR #355, opened 2024-09-24
Implement solution (also push migrations like proof of working db)

dimak20:


dimak20:
image
image
image
image
image
image

db/models.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+class DBBook(Base):
+    __tablename__ = "book"
+
+    id = Column(Integer, primary_key=True, index=True)
+    title = Column(String(255), nullable=False)
+    summary = Column(String(255), nullable=False)
+    publication_date = Column(DATE, nullable=False)
+    author_id = Column(Integer, ForeignKey("author.id"))
```
Oleksl888, line 24: I think would be logical to make title and author_id unique together

Oleksl888 (approved):
It is ok - but few notes. ( I know that you followed the requirements).
1. Routers should be placed in separate file and plugged into main app.
2. Project structure could be improved: e.g. create package `library` and put models, schemas, routers and crud
3. In crud functions are repeating, it can be refactored and parametrized. e.g. instead of create_book, create_author etc have method create_entry that will accept model and schema along with data and session. Once parametrized you can call this function on different endpoints with correct arguments and you will not need to write extra functions in crud to suport new models

PR #356, opened 2024-09-25
 Solution FastApi

Ramzes525:


NazarSheremet (changes requested):
Remove from PR useless files
image

app/crud.py
NazarSheremet, line 1: This files look like models, not crud

NazarSheremet (changes requested):
Add file with crud methods

NazarSheremet (approved):
Good job!

PR #548, opened 2025-12-23
Almost finish

Leo9siy:


routers/author_routers.py
```diff
@@ -0,0 +1,57 @@
...
+    return AuthorListSchema(authors=select_authors(db, skip, limit))
+
+
+@author_router.get(
+    path="/authors/{author_id}/",
+    response_model=AuthorResponseSchema,
+    status_code=200,
+    description="Retrieve a single author by ID."
+)
+def get_author_by_id(
```
DaTrEvTeR, line 49: The get_by_id function should be implemented in CRUD. This way, we separate the logic for working with the database, preparing responses, and (in real projects) the service layer (our business logic). Each component doesn't need to know about the others to work, which helps with code growth.

crud.py
```diff
@@ -0,0 +1,51 @@
...
+    )
+
+    db.add(book)
+    db.commit()
+    db.refresh(book)
+
+    return book
+
+
+def select_books(db: Session, limit: int = 10, skip: int = 0, author_id: int = None):
```
DaTrEvTeR: Better rename this func into select_author_books or something else

crud.py
```diff
@@ -0,0 +1,51 @@
...
+    db.add(author)
+    db.commit()
+    db.refresh(author)
+
+    return author
+
+
+def select_authors(db: Session, skip: int = 0, limit: int = 10):
+    authors = db.query(AuthorModel).offset(skip).limit(limit).all()
+    authors = [AuthorResponseSchema.model_validate(author) for author in authors]
```
DaTrEvTeR: It's a crud service. Service to work with db. Not for response preparing(parse models to schema)

schemas.py
```diff
@@ -0,0 +1,39 @@
...
+    model_config = {"from_attributes": True}
+
+
+class AuthorListSchema(BaseModel):
+    authors: list[AuthorResponseSchema]
+
+    model_config = {"from_attributes": True}
+
+
+class BookCreateSchema(BaseModel):
```
DaTrEvTeR: In all your models, the only required field is the book.title. All other fields can be empty, and you don't include it in your schemas. This could lead to errors in your API. You need to make all fields nullable=False or use oprional typing for fields in schemas: date | None or Optional[date](with import Optional from typing)

DaTrEvTeR (changes requested):

vsmutok (changes requested):
It’s bad practice to keep everything in the project root.
Create separate folders for routes and for the database schemas, and move the modules there.

vsmutok (approved):
Good Job!

DaTrEvTeR (approved):

PR #573, opened 2026-01-25
Develop

Paul-Starodub:


models.py
```diff
@@ -0,0 +1,44 @@
...
+    summary: Mapped[str]
+    publication_date: Mapped[date]
+    author_id: Mapped[int] = mapped_column(ForeignKey("authors.id"))
+    author: Mapped["Author"] = relationship(back_populates="books")
+
+    def __repr__(self) -> str:
+        return self.title
+
+
+# example of one to one relationship
```
vsmutok, line 38: What are these kinds of comments for here?

vsmutok (changes requested):

models.py
Paul-Starodub: Just for my practice
This task was completed two years ago.

crud.py
DaTrEvTeR, line 1: It's better to create a class named, for example, `AuthorRepository` and add methods to it with the `@classmethod `decorator instead of using simple functions. You can also use the `singleton` pattern for this class or simply create a single object and import it into other modules (as you did with settings).

DaTrEvTeR (changes requested):
You did a great job! But there are still a few areas that need improvement.

crud.py
Paul-Starodub: Done

vsmutok (approved):
Great!

DaTrEvTeR (approved):

PR #723, opened 2026-08-10
solution 1

pprykhodko:


library.sqlite3
Sn1kls, line 1: The database file shouldn't be tracked in git – it's a build artifact, and every run produces a different binary diff. Please remove it from the index:
`git rm --cached library.sqlite3`


Also update `.gitignore` from `db.sqlite3` to `*.sqlite3` so any local DB is ignored.

main.py
```diff
@@ -0,0 +1,130 @@
...
+import crud
+import schemas
+from database import Base, SessionLocal, engine
+
+
+Base.metadata.create_all(bind=engine)
+app = FastAPI()
+
+
+def get_database() -> Session:
```
Sn1kls: `get_database()` is a generator, not a function returning a `Session`, so the
annotation is misleading:
```
    from collections.abc import Generator

    def get_database() -> Generator[Session, None, None]:
```

main.py
```diff
@@ -0,0 +1,130 @@
...
+            detail="Author not found"
+        )
+
+    return database_author
+
+
+@app.post("/authors/", response_model=schemas.AuthorRead, status_code=201)
+def create_author(
+        author: schemas.AuthorCreate,
+        database: Session = Depends(get_database),
```
Sn1kls: Here you use `database: Session = Depends(get_database)`, but in `read_authors` and `read_books` you use `Annotated[Session, Depends(get_database)]`. Pick one style – `Annotated` is the recommended one – and apply it everywhere (same in `create_book`, line 113).

alembic/versions/e45558ad1723_second_migartion.py
```diff
@@ -0,0 +1,32 @@
...
+    # ### commands auto generated by Alembic - please adjust! ###
+    pass
+    # ### end Alembic commands ###
+
+
+def downgrade() -> None:
+    """Downgrade schema."""
+    # ### commands auto generated by Alembic - please adjust! ###
+    pass
+    # ### end Alembic commands ###
```
Sn1kls: Optional / for the future: this revision and the next one are empty – `upgrade()` and `downgrade()` are just `pass`. That happens when `alembic revision --autogenerate` runs while the models haven't changed. It's harmless (the chain still upgrades fine), but it's worth checking the generated file before committing it. In a fresh solution like this one a single initial migration would be enough.

Also a small typo in the file names: "migartion" –> "migration".

If you do decide to squash them, delete your local `library.sqlite3` too – otherwise Alembic will complain that it can't locate revision `6ae55eb29b76`, which is stamped in your current DB.

Sn1kls (changes requested):

Sn1kls (approved):
Approve! Good work)

PR #726, opened 2026-08-12
Implement Library Management API with FastAPI

omerlenko:


schemas.py
```diff
@@ -0,0 +1,34 @@
...
+
+class AuthorRead(AuthorBase):
+    id: int
+    model_config = ConfigDict(from_attributes=True)
+
+
+class BookBase(BaseModel):
+    title: Annotated[str, Field(max_length=255)]
+    summary: Annotated[str, Field(max_length=511)]
+    publication_date: datetime
```
Sn1kls: The model and migration now use date (good), but the schema still uses datetime, so the API serializes `publication_date` as `1997-06-26T00:00:00` instead of `1997-06-26`. Change it to date:

```
from datetime import date
...
publication_date: date
```

Sn1kls (changes requested):

Sn1kls (approved):
Approved. Clean work!)

